PR #6, opened 2025-10-28
Some UI Cleanup Items

declan-scale:
## Changes
- Update the chat scroll when the number of messagePairs is updated -> only updated on user message send
- Remove the old toast and claude files
- Format the scroll to bottom button to be above the chat, not absolutely from the bottom
- Add a better custom toast with a title and optional message
- Show toast on message send or task creation failure
- Make the whole dropdown for reasoning / tool calls clickable instead of just chevron
- Update the scroll button visibility on every task change (won't be visible randomly when there is no scroll)

### Scroll button
image

### Toast
image


MichaelSun48 (approved):
hell yeah

PR #7, opened 2025-10-28
Add cmd+k shortcut

MichaelSun48:
Adds a cmd+k shortcut for creating a new task. 

Adds an indicator to the new chat button: 

image

### DIFF
--- a/agentex-ui/components/agentex/task-sidebar.tsx
+++ b/agentex-ui/components/agentex/task-sidebar.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 
 import Image from 'next/image';
 import { useRouter } from 'next/navigation';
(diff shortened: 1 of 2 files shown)


agentex-ui/components/agentex/task-sidebar.tsx
```diff
@@ -99,6 +100,22 @@ export function TaskSidebar({
     setIsCollapsed(prev => !prev);
   }, []);
 
+  // Global keyboard shortcut: cmd + k for new chat
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
```
declan-scale: Would this be better in `main-content`?

MichaelSun48: Hmm yeah actually I put it here initially for parity with the new chat button, but I guess if we ever choose to unrender the task-sidebar, we'd probably still want to preserve this behavior.

MichaelSun48: Will move it, good suggestion!

PR #14, opened 2025-10-29
Fix use infinite tasks hook

MichaelSun48:
Removes the useTasks hook (was not in use) and fixes the useCreateTask hook to respect the InfiniteData type that is returned from the useInfiniteQuery hook 

This caused a bug on the frontend where trying to create a task after fetching all tasks would cause an error


declan-scale (approved):
Good catch!

PR #15, opened 2025-10-29
Restructure agentex-ui components

MichaelSun48:
**Dependent on #14 **

This PR makes a bunch of DevEX improvements to the agentex-ui components, mostly around improving component naming and organization. Rather than enumerating them here, I'll annotate the diff which is probably easier to follow. 

 No functionality should be touched — if you notice anything, please let me know.

### DIFF
--- a/agentex-ui/components/agentex/task-sidebar.tsx
+++ b/agentex-ui/components/agentex/task-sidebar.tsx
@@ -112,7 +112,7 @@ function TaskButton({ task }: TaskButtonProps) {
           )}
         >
           {createdAtString}
-          {task.created_at && ' • '}
+          {task.agents && task.agents.length > 0 && task.created_at && ' • '}
           {agentsString}
         </div>
       </Button>


agentex-ui/app/agentex-ui-root.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <Suspense fallback={<div>Loading...</div>}>
+          <div className="fixed inset-0 flex w-full">
+            <TaskSidebar />
+            <PrimaryContent
+              isTracesSidebarOpen={isTracesSidebarOpen}
+              toggleTracesSidebar={() =>
+                setIsTracesSidebarOpen(!isTracesSidebarOpen)
+              }
+            />
+            <TracesSidebar isOpen={isTracesSidebarOpen} />
```
MichaelSun48, line 82: Note here — hopefully this is much more intuitive and conveys the three part page structure we have

agentex-ui/app/primary-content.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+import { ChatView } from './chat-view';
+import { HomeView } from './home-view';
+
+interface ContentAreaProps {
+  isTracesSidebarOpen: boolean;
+  toggleTracesSidebar: () => void;
+}
+
+export function PrimaryContent({
```
MichaelSun48, line 24: This component is the middle third that contains the chat and home screen. Honestly, this one is still a little too big and could be pared down even more, but for teh sake of keeping this PR reasonable, I'll make those in a follow up.

agentex-ui/app/primary-content.tsx
```diff
@@ -0,0 +1,154 @@
...
+            taskId={taskID}
+            isTracesSidebarOpen={isTracesSidebarOpen}
+            toggleTracesSidebar={toggleTracesSidebar}
+            agents={agents}
+            onAgentChange={handleSelectAgent}
+          />
+        </motion.div>
+      )}
+
+      {taskID ? <ChatView taskID={taskID} /> : <HomeView />}
```
MichaelSun48, line 108: Note here — these two components definitely should be in their own files

agentex-ui/app/primary-content.tsx
```diff
@@ -0,0 +1,154 @@
...
+                animate={{ y: 0, opacity: 1 }}
+                exit={{ y: 30, opacity: 0 }}
+                transition={{
+                  duration: 0.2,
+                  type: 'spring',
+                  stiffness: 300,
+                  damping: 35,
+                  mass: 0.8,
+                }}
+              >
```
MichaelSun48, line 137: This is definitely my fault, but in a future PR I'm going to move these motion.div into the components, rather than having them in the parent

agentex-ui/components/agentex/task-sidebar.tsx
```diff
@@ -19,21 +19,32 @@ import { useAgentexClient } from '@/components/providers';
...
 type TaskButtonProps = {
   task: TaskListResponse.TaskListResponseItem;
-  selectedTaskID: TaskListResponse.TaskListResponseItem['id'] | null;
-  selectTask: (
-    taskID: TaskListResponse.TaskListResponseItem['id'] | null
-  ) => void;
 };
 
-function TaskButton({ task, selectedTaskID, selectTask }: TaskButtonProps) {
+function TaskButton({ task }: TaskButtonProps) {
```
MichaelSun48, line 35: Common theme in this PR is that I'm trying to get rid of anywhere we pass taskId or agentName through props and instead let any component that needs it get it via useSafeSearchParams hook.

agentex-ui/components/agentex/task-sidebar.tsx
```diff
@@ -101,27 +112,16 @@ function TaskButton({ task, selectedTaskID, selectTask }: TaskButtonProps) {
           )}
         >
           {createdAtString}
-          {task.agents && task.agents.length > 0 && task.created_at && ' • '}
+          {task.created_at && ' • '}
```
MichaelSun48: if tasks.agent is nothing, the agents string still exists and says "No Agents". I was seeing this: 

Image

agentex-ui/components/agentex/task-sidebar.tsx
```diff
@@ -162,16 +162,11 @@ export function TaskSidebar({
...
-    [onSelectTask]
-  );
-
   const handleNewChat = useCallback(() => {
-    onSelectTask(null);
-  }, [onSelectTask]);
+    updateParams({
+      [SearchParamKey.TASK_ID]: null,
+    });
+  }, [updateParams]);
```
MichaelSun48, line 169: Note here — another example of letting a component consume the useSearchParams hook rather than receiving it as props.

agentex-ui/components/agentex/task-sidebar.tsx
```diff
@@ -101,27 +112,16 @@ function TaskButton({ task, selectedTaskID, selectTask }: TaskButtonProps) {
           )}
         >
           {createdAtString}
-          {task.agents && task.agents.length > 0 && task.created_at && ' • '}
+          {' • '}
```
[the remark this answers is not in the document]
declan-scale: I would do the suggestion

agentex-ui/components/agentex/task-sidebar.tsx
declan-scale: This is interesting. Initially `No date` was put there so we could set the block to invisible and maintain the height of the element. Probably better to just set the createdAtString to '' and the agentsString to '' instead of the dummy strings that exist currently. Then we could just manually set the height along with the invisible tailwind class above.

agentex-ui/components/agentex/task-sidebar.tsx
declan-scale, line 35: Does it make sense to pop this out into its own TaskSidebarItem.tsx file?

MichaelSun48: Yeah definitely — I'll include that in a follow up PR

PR #29, opened 2025-10-31
Async streaming tasks fix

declan-scale:
## Changes
- Make GET `/tasks` order by desc instead of asc (got reverted somehow)
- Remove redundant QueryProvider (already in `layout.tsx`)
- Default stale time is infinity to limit redundant calls (open to being a but smarter about this)
- Added `updateTaskInInfiniteQuery` to update the relevant caches both `onTaskChange` for the SSE side and when we create a task
    - This was the cause of a couple issues:
        1. the SSE stream being cut on the FE because of rerenders
        2. fetching the whole task list after landing on a specific task page would throw an error
- Updated some query keys to distinguish between individual tasks and all

### DIFF
--- a/agentex-ui/components/providers/task-provider.tsx
+++ b/agentex-ui/components/providers/task-provider.tsx
@@ -23,7 +23,8 @@ export function TaskProvider({
   useTaskSubscription({
     agentexClient,
     taskId,
-    enabled: !!taskId && agent?.acp_type === 'agentic',
+    agentName: agentName || '',
+    enabled: !!taskId && (agent?.acp_type === 'agentic' || !agentName),
   });
 
   return <>{children}</>;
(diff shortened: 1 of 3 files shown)


agentex-ui/app/agentex-ui-root.tsx
```diff
@@ -65,7 +64,7 @@ export function AgentexUIRoot({
   }, [handleSelectTask]);
 
   return (
-    <QueryProvider>
+    <>
```
MichaelSun48, line 67: Won't all of the useQuery's error out without this

MichaelSun48 (approved):
Just that one comment, but I could totally just be missing something

agentex-ui/app/agentex-ui-root.tsx
declan-scale: QueryProvider is already supplied in layout.tsx.

PR #30, opened 2025-10-31
Agentex-UI cleanup

declan-scale:
Worked to get the code base cleaned up. If you follow the changes commit by that would probably be best as the changes are iterative. I grouped each change into a commit

## Changes
- Change the Response component -> MarkdownResponse, move it into the agentex directory and delete the ai-elements
    - Remove CodeBlock (was unused)
- Remove all the unused packages
- Use types for props instead of interfaces (we were using a bit of both)
- Memoize components with the following pattern:
    - declare a functional component called `ComponentNameImpl`
    - set `ComponentName = memo(ComponentNameImpl)`
    - `export { ComponentName }`
- Removed more unused packages after removing some unused components (the old input component)
- Remove the unused SubmitButton and FileUploadButton
- Only export types used in other files
- Move utils and `JsonValue` into the lib directory
- Use absolute import paths (as opposed to relative), rename components with the words Memoized or Component in the name
- Add vitest for the utils

### DIFF
--- a/agentex-ui/eslint.config.js
+++ b/agentex-ui/eslint.config.js
@@ -15,6 +15,8 @@ module.exports = [
       'postcss.config.mjs',
       'next.config.ts',
       'next-env.d.ts',
+      'vitest.config.mts',
+      'vitest.setup.ts',
     ],
   },
   ...compat.extends('next/core-web-vitals', 'next/typescript', 'prettier'),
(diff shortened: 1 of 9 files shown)


agentex-ui/eslint.config.js
```diff
@@ -15,6 +15,8 @@ module.exports = [
       'postcss.config.mjs',
       'next.config.ts',
       'next-env.d.ts',
+      'vitest.config.mts',
+      'vitest.setup.ts',
```
MichaelSun48, line 19: 🙏

PR #49, opened 2025-11-05
Add release workflow

RoxyFarhad:


.github/workflows/build-agentex.yml
```diff
@@ -0,0 +1,109 @@
...
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  AGENTEX_SERVER_IMAGE_NAME: agentex
+  AGENTEX_AUTH_IMAGE_NAME: agentex-auth
```
jasonyang101, line 15: should remove this

vishwas896:
System is not functional correction, hence requesting for pull requesting activation during the intervals, pas or users get bugs.

PR #51, opened 2025-11-05
only run integration tests when changes are made to agentex/

MichaelSun48:
Only runs the integration tests, unit test , and docs tests for a PR if there are changes made to agentex/. There's not need to waste compute on this for ui changes, for example 

(notice how these unit tests did not run for this PR)


.github/workflows/integration-tests.yml
MichaelSun48, line 1: We should probably move this action to the ci.yml file

PR #56, opened 2025-11-05
Update healthcheck to confirm agent ID if available

smoreinis:
For local development, this means that healthcheck will be able to distinguish which agent is running if multiple ones are registered to the same ACP URL

This also adds back in the script to ensure all agent workflows are kicked off on local startup during local development

### DIFF
--- a/agentex/src/temporal/run_healthcheck_workflow.py
+++ b/agentex/src/temporal/run_healthcheck_workflow.py
@@ -11,6 +11,7 @@ from src.config.dependencies import (
     database_async_read_write_session_maker,
 )
 from src.config.environment_variables import EnvironmentVariables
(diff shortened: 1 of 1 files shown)


declan-scale:
@smoreinis am I missing something or is there no logic for the id check? Maybe it got lost in a merge?

smoreinis:
Yeah, it tries to kick off a workflow for every registered agent but if that already happened before (and the workflow was still running before the container got terminated) then we'll get the log message above.
I just updated it to only try kicking off a workflow for those agents that are READY, so if you have the SDK change in then it should be marking the rest of the agents UNHEALTHY and only trying to start one workflow on startup (per ACP URL).

PR #66, opened 2025-11-07
Clean up primary content and fix scrollToBottom button not appearing

MichaelSun48:
This PR makes a couple changes to the PrimaryContent component: 

1. Fixes a bug where the scroll to bottom button would not appear. This was because we were not attaching the scroll container ref to the scroll container lol 
2. Moves the task header component directly into the chat view. This component only renders when the chat view exists, which is why this makes sense. 
3. Moves the scroll to bottom button to a separate component, mostly to make the primary content component cleaner 
4. Removes some unnecessary wrapper divs in a few places

### DIFF
--- a/agentex-ui/components/task-messages/task-message-scroll-container.tsx
+++ b/agentex-ui/components/task-messages/task-message-scroll-container.tsx
@@ -17,10 +17,7 @@ const TaskMessageScrollContainer = forwardRef<
   return (
     <div
       ref={ref}
-      className={cn(
-        'flex w-full flex-col gap-4 overflow-y-auto px-4 pt-4',
-        className
-      )}
+      className={cn('flex w-full flex-col gap-4 px-4 pt-4', className)}
       style={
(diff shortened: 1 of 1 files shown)


agentex-ui/components/task-messages/task-messages.tsx
```diff
@@ -114,7 +115,9 @@ function TaskMessagesImpl({ taskId }: TaskMessagesProps) {
         while (element) {
           const overflowY = window.getComputedStyle(element).overflowY;
           if (overflowY === 'auto' || overflowY === 'scroll') {
-            setContainerHeight(element.clientHeight);
+            setContainerHeight(
+              element.clientHeight - (headerRef.current?.clientHeight ?? 0)
+            );
```
MichaelSun48, line 120: Kinda annoying part about moving the header directly into the chat view is that we have to adjust this min height calculation to include the height of the header, otherwise its too big and the first chat message appears under the header

PR #83, opened 2025-11-26
Add high limit to bulk message endpoint

MichaelSun48:
Replaces the agentex-sdk's implementation of subscribeTaskState with a custom, near identical version that includes a 1000 limit on the messages endpoint. 

This is intended to be a stop gap solution to an issue where only the last 50 messages appear in the UI. The long term fix will be to add an order_by field to our bulk get endpoints and changing the typescript sdk to order the messages by desc. This PR will be effectively reverted once that is completed.

### DIFF
--- a/agentex-ui/hooks/custom-subscribe-task-state.tsx
+++ b/agentex-ui/hooks/custom-subscribe-task-state.tsx
@@ -105,8 +105,6 @@ export async function subscribeTaskState(
     taskStreamReconnectPolicy?.exponentialBackoffBase ?? 2;
 
   // current subscription state
-  let _task: Task | null = null;
-  let _agents: Agent[] | null = null;
   let messages: TaskMessage[] | null = null;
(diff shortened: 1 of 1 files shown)


agentex-ui/hooks/use-task-subscription.ts
```diff
@@ -1,13 +1,15 @@
...
 import { InfiniteData, useQueryClient } from '@tanstack/react-query';
-import { subscribeTaskState } from 'agentex/lib';
+// import { subscribeTaskState } from 'agentex/lib';
 
 import { updateTaskInInfiniteQuery } from '@/hooks/use-create-task';
 import { taskMessagesKeys } from '@/hooks/use-task-messages';
 import type { TaskMessagesData } from '@/hooks/use-task-messages';
 import { tasksKeys } from '@/hooks/use-tasks';
 
+import { subscribeTaskState } from './custom-subscribe-task-state';
```
declan-scale: [nit] use absolute path

PR #84, opened 2025-11-26
Add orderby field to postgres endpoints

MichaelSun48:
Adds an order_by field to the following model's get endpoints: 
1. Tasks 
2. Agents
3. Spans 
4. Agent Task Tracker 
5. Deployment History

I've only made this change to the postgres models since it was easy to just expose the order_by field in the ORM up to the API layer. Doing it for the mongoDB objects might be _slightly_ more complex and should probably come in a separate PR.


declan-scale (approved):
Is there any validation that the order_by field is an actual attribute of the object?

MichaelSun48:
> Is there any validation that the order_by field is an actual attribute of the object? 

@declan-scale Hmm not at the API layer, no. I would assume it would just fail at the ORM level, but yeah I do think it would make sense to add this at an API layer

PR #85, opened 2025-12-01
Add order by and order direction field to mongodb stores

MichaelSun48:
Adds the order by and order direction fields to the messages and states objects, along with some tests.


agentex/src/adapters/crud_store/adapter_mongodb.py
```diff
@@ -525,8 +527,20 @@ async def list(
...
+            direction = (
+                pymongo.DESCENDING
+                if order_direction and order_direction.lower() == "desc"
+                else pymongo.ASCENDING
+            )
+            sort_list.append((order_by, direction))
+
+        # Always use _id as tiebreaker
+        sort_list.append(("_id", pymongo.ASCENDING))
+        cursor = cursor.sort(sort_list)
```
MichaelSun48, line 542: This is the critical change in our MongoDB ORM logic

smoreinis (approved):

PR #92, opened 2025-12-07
feat: cursor-based pagination for messages API

gabrycina:
## Summary

Implements cursor-based pagination for the `/messages` endpoint to enable infinite scroll in the UI.

### Breaking Change

`GET /messages` now returns `PaginatedMessagesResponse` instead of raw array:
```json
{ "data": [...], "next_cursor": "...", "has_more": true }
```

### Backend Changes
- Add `cursor` and `direction` query params to `/messages`
- Return `{data, next_cursor, has_more}` response format
- Add `pagination.py` with cursor encode/decode utilities
- Add `find_by_field_with_cursor()` to MongoDB adapter

### Frontend Changes
- Replace `useTaskMessages` with `useInfiniteTaskMessages` hook
- Implement infinite scroll with IntersectionObserver
- Add scroll position preservation when loading older messages
- Fix Thinking indicator timing
- Remove deprecated `custom-subscribe-task-state.tsx`

### Related PRs
- TypeScript SDK: TBD
- Python SDK: https://github.com/scaleapi/scale-agentex-python/pull/222

### Suggested Deployment Order
1. Deploy this PR (backend)
2. Publish TypeScript SDK
3. Publish Python SDK
4. Deploy frontend (after updating SDK version)


agentex/src/adapters/crud_store/adapter_mongodb.py
```diff
@@ -605,6 +605,85 @@ async def find_by_field(
...
+                    cursor_object_id = ObjectId(cursor_id)
+                except Exception:
+                    cursor_object_id = cursor_id
+
+                cursor_doc = self.collection.find_one({"_id": cursor_object_id})
+                if cursor_doc and "created_at" in cursor_doc:
+                    cursor_timestamp = cursor_doc["created_at"]
+                    if before_id:
+                        # Get documents with created_at < cursor's created_at
+                        query["created_at"] = {"$lt": cursor_timestamp}
```
smoreinis: Since this uses lt / gt, do we assume that there won't be another message with the same timestamp? (presumably this shouldn't happen given the granularity for the timestamps)

gabrycina: Agree, addressed with compound comparison (id)

PR #94, opened 2025-12-09
Feature/cursor pagination messages UI

gabrycina:


agentex-ui/pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+overrides:
+  agentex: link:../../scale-agentex-typescript/dist
```
MichaelSun48: Pls remove this an the pnpm-lock file (might need to rebase on main to see these files in .gitignore)

PR #97, opened 2025-12-16
feat: adding filters to the ListMessages

RoxyFarhad:
This allows a user to filter TaskMessages based on any data in the TaskMessageResponse. 

The filter is a JSON encoded query parameter that is supplied as part of the get request and the DB does **partial matching** to find relevant matching entries while also adhering to cursor based pagination.

### DIFF
--- a/agentex/src/domain/entities/task_messages.py
+++ b/agentex/src/domain/entities/task_messages.py
@@ -247,12 +247,6 @@ class TaskMessageEntityFilter(BaseModel):
     streaming_status: Literal["IN_PROGRESS", "DONE"] | None = Field(
         None, description="Filter by streaming status"
     )
-    created_at: datetime | None = Field(
-        None, description="Filter by message creation timestamp"
-    )
-    updated_at: datetime | None = Field(
-        None, description="Filter by message last update timestamp"
-    )
 
 
 def convert_task_message_to_entity(message: TaskMessage) -> TaskMessageEntity:
--- a/agentex/src/utils/model_utils.py
+++ b/agentex/src/utils/model_utils.py
@@ -79,42 +79,6 @@ class BaseModel(PydanticBaseModel):
         return value
 
 
-def exclude_fields(
-    entity_model: type[BaseModel],
-    exclude_fields: set[str],
-    model_name_suffix: str = "Filter",
-) -> type[BaseModel]:
-    """
-    Create a filter model by excluding specific fields from an entity model.
-
-    Args:
-        entity_model: The source entity model to base the filter on
-        exclude_fields: Fields to exclude from the filter model (defaults to {"id"})
-        model_name_suffix: Suffix for the generated model name (defaults to "Filter")
-
-    Returns:
-        A new Pydantic model class with specified fields excluded
-
-    Example:
-        TaskMessageFilter = create_filter_model(TaskMessageEntity)
-        UserFilter = create_filter_model(UserEntity, {"id", "password_hash"})
-    """
-
-    # Extract fields from source model, excluding specified ones
-    fields = {}
-    for field_name, field_info in entity_model.model_fields.items():
-        if field_name not in exclude_fields:
-            fields[field_name] = (field_info.annotation, field_info)
-
-    # Create new model class dynamically
-    filter_model_name = f"{entity_model.__name__}{model_name_suffix}"
-    return create_model(
-        filter_model_name,
-        __base__=BaseModel,  # Use our custom BaseModel
-        **fields,
-    )
-
-
 def make_optional(
     entity_model: type[BaseModel], model_name_suffix: str = "Optional"
 ) -> type[BaseModel]:


agentex/src/adapters/crud_store/adapter_mongodb.py
```diff
@@ -659,25 +667,27 @@ async def find_by_field_with_cursor(
...
+                        # Get documents where (for descending created_at, ascending _id sort):
+                        # - created_at < cursor_timestamp (strictly older), OR
+                        # - created_at == cursor_timestamp AND _id > cursor_id (same timestamp,
+                        #   but later in ascending _id order, i.e., comes after cursor in results)
                         query["$or"] = [
                             {"created_at": {"$lt": cursor_timestamp}},
                             {
                                 "created_at": cursor_timestamp,
-                                "_id": {"$lt": cursor_object_id},
+                                "_id": {"$gt": cursor_object_id},
```
RoxyFarhad, line 678: This was a bug that was returning duplicate messages based on the ordering. 
```
Say we have 6 messages created rapidly (same millisecond timestamp):

  Example Data

  | Position | created_at | _id |
  |----------|------------|-----|
  | 1        | 10:00:00   | aaa |
  | 2        | 10:00:00   | bbb |
  | 3        | 10:00:00   | ccc |
  | 4        | 10:00:00   | ddd |
  | 5        | 09:59:59   | zzz |
  | 6        | 09:59:58   | yyy |

```
In this case, we are looking for items in DESC order, and using the object_id as the cursor object. 

That means that after:
Page 1 (limit=3, no cursor)
Returns positions 1-3: [aaa, bbb, ccc]

Cursor is set to the last item: ccc at 10:00:00

That means that the next cursor needs to look at the items AFTER ccc, not before. What we had before would return `bbb` as the first item which was causing duplicate messages to show.

agentex/src/adapters/crud_store/adapter_mongodb.py
```diff
@@ -659,25 +667,27 @@ async def find_by_field_with_cursor(
...
+                        # Get documents where (for descending created_at, ascending _id sort):
+                        # - created_at > cursor_timestamp (strictly newer), OR
+                        # - created_at == cursor_timestamp AND _id < cursor_id (same timestamp,
+                        #   but earlier in ascending _id order, i.e., comes before cursor in results)
                         query["$or"] = [
                             {"created_at": {"$gt": cursor_timestamp}},
                             {
                                 "created_at": cursor_timestamp,
-                                "_id": {"$gt": cursor_object_id},
+                                "_id": {"$lt": cursor_object_id},
```
RoxyFarhad, line 690: Relates to the same bug above ^  but in the reverse direction. 

if after_id, is set, it means that we want to find documents AFTER the timestamp. that means on a tie-break, we need to find the earlier objectIDs because the later ones would have already been shown in the previous cursor.

agentex/src/domain/entities/task_messages.py
```diff
@@ -81,6 +81,9 @@ class TextContentEntity(BaseTaskMessageContentEntity):
     )
 
 
+OptionalTextContentEntity = make_optional(TextContentEntity)
```
RoxyFarhad, line 84: We essentially take all the Content objects and make each field optional. This then becomes the query input params without having to manually update this object as we add to it.

agentex/src/domain/repositories/task_message_repository.py
```diff
@@ -36,6 +36,15 @@ class TaskMessageRepository(MongoDBCRUDRepository[TaskMessageEntity]):
             "name": "task_id_streaming_status_idx",
             "description": "Index for streaming status queries during streaming operations",
         },
+        {
+            "keys": [
+                ("task_id", pymongo.ASCENDING),
+                ("content.type", pymongo.ASCENDING),
```
RoxyFarhad, line 42: this adds an index at least to filter by the content types. we may want to enforce this at the data layer otherwise there is a world in which this could start not being performant

agentex/src/utils/model_utils.py
```diff
@@ -77,3 +77,86 @@ def validate_to_json(cls, value):
         if isinstance(value, str):
             return cls(**json.loads(value))
         return value
+
+
+def exclude_fields(
```
smoreinis: not sure if it's obvious, but where does this get used in this PR?

RoxyFarhad: its acc not anymore! good call out -- will remove sorry

agentex/src/domain/entities/task_messages.py
```diff
@@ -216,6 +238,23 @@ class TaskMessageEntity(BaseModel):
...
+class TaskMessageEntityFilter(BaseModel):
+    """Filter model for TaskMessage - all fields optional for flexible filtering."""
+
+    content: OptionalTaskMessageContentEntity | None = Field(
+        None, description="Filter by message content"
+    )
+    streaming_status: Literal["IN_PROGRESS", "DONE"] | None = Field(
+        None, description="Filter by streaming status"
+    )
+    created_at: datetime | None = Field(
```
smoreinis: Just to confirm, this is meant to support equality matching on the exact creation / updated timestamp? Or is it to prepare for supporting filtering for messages older / newer than a certain timestamp? 
Is it worth adding a basic test case for this as well?

RoxyFarhad: i think im going to actually remove this for now instead of supporting it -- feel like we may need to do some additional indexing here on the DB / do some things with cursors

gabrycina (approved):

agentex/src/domain/entities/task_messages.py
```diff
@@ -126,16 +138,26 @@ class DataContentEntity(BaseTaskMessageContentEntity):
     data: dict[str, Any] = Field(..., description="The contents of the data message.")
 
 
+OptionalDataContentEntity = make_optional(DataContentEntity)
+
 TaskMessageContentEntity = Annotated[
     TextContentEntity
-    | ReasoningContentEntity
```
danielmillerp, line 131: any reason this was removed?

PR #100, opened 2025-12-17
Add negation to filtering and better schemas to docs

RoxyFarhad:
The first pass only supported inclusionary filtering. This PR adds a new field to the API that says: "exclude" which defaults to false. If you set this value to be true, it will find documents that negate this query.

### DIFF
--- a/agentex/src/domain/use_cases/messages_use_case.py
+++ b/agentex/src/domain/use_cases/messages_use_case.py
@@ -54,10 +54,7 @@ def convert_filters_to_mongodb_query(
     # Build include query (OR'd together)
     if include_filters:
         converted = [_convert_single_filter(f) for f in include_filters]
-        if len(converted) == 1:
-            include_query = converted[0]
-        else:
-            include_query = {"$or": converted}
+        include_query = {"$or": converted}
 
     # Build exclude query (OR'd together, then $nor)
     if exclude_filters:


agentex/src/domain/use_cases/messages_use_case.py
```diff
@@ -10,49 +10,69 @@
...
+    if include_filters:
+        converted = [_convert_single_filter(f) for f in include_filters]
+        if len(converted) == 1:
+            include_query = converted[0]
+        else:
+            include_query = {"$or": converted}
+
+    # Build exclude query (OR'd together, then $nor)
+    if exclude_filters:
+        converted = [_convert_single_filter(f) for f in exclude_filters]
```
smoreinis, line 61: what if there's only one exclude_filter, should there be a case where it's turned into a `{"$not": converted[0]}`?
Conversely, if a single clause nor works, wondering why we avoid using a single clause or above.

agentex/src/domain/entities/task_messages.py
```diff
@@ -239,14 +239,24 @@ class TaskMessageEntity(BaseModel):
 
 
 class TaskMessageEntityFilter(BaseModel):
-    """Filter model for TaskMessage - all fields optional for flexible filtering."""
+    """Filter model for TaskMessage - all fields optional for flexible filtering.
+
+    The `exclude` field determines whether this filter is inclusionary or exclusionary.
+    When multiple filters are provided:
+    - Inclusionary filters (exclude=False) are OR'd together
+    - Exclusionary filters (exclude=True) are OR'd together and negated with $nor
```
smoreinis, line 247: is it fair to just say here that exclusionary filters are NOR'd together?

RoxyFarhad: yes exactly, this is the brute force approach.

agentex/src/domain/entities/task_messages.py
```diff
@@ -239,14 +239,24 @@ class TaskMessageEntity(BaseModel):
...
+    - The two groups are AND'd: (include1 OR include2) AND NOT (exclude1 OR exclude2)
```
smoreinis, line 248: This makes sense to me, but it also feels a little off that it doesn't match the actual logic which uses NORs. Thoughts on actually having this description match the implementation and combine the exclusionary clauses via `{"$not": {"$or": [exclude1, exclude2]}}`?

agentex/src/domain/use_cases/messages_use_case.py
RoxyFarhad: yeah good call out for the single nor clause, will update ^ My op is just simplifying to always use the or clause even if the length is 1 -- mongoDB has a query optimizer so there isn't any perf difference using either way also

agentex/src/domain/entities/task_messages.py
RoxyFarhad: `{"$not": {"$or": [exclude1, exclude2]}}` isn't valid for mongo as it only supports a single field expression, which is why the `$nor` is needed. 

Ill update the description though

PR #109, opened 2025-12-22
Add readonly database session maker for read replica support

smoreinis:
## Summary

- Add `READ_ONLY_DATABASE_URL` environment variable for connecting to a read replica
- Activate `DDatabaseAsyncReadOnlySessionMaker` dependency for FastAPI injection
- Falls back to `DATABASE_URL` when no replica URL is configured (seamless dev experience)
- Update all repositories to accept both RW and RO session makers via FastAPI DI
- Add DB-enforced read-only sessions in integration tests using `SET TRANSACTION READ ONLY`
- Add ruff linting commands to CLAUDE.md

Closes AGX1-22

## Test plan

- [x] Unit tests pass (`make test-unit`) - 184 tests
- [x] Integration tests pass (`make test-integration`) - 66 tests
- [x] Ruff lint checks pass

### DIFF
--- a/agentex/tests/integration/fixtures/integration_client.py
+++ b/agentex/tests/integration/fixtures/integration_client.py
@@ -229,9 +229,25 @@ async def isolated_repositories(isolated_test_schema):
     mongodb_database = isolated_test_schema["mongodb_database"]
     redis_client = isolated_test_schema["redis_client"]
 
-    # Create async session factory for PostgreSQL
-    async_session_factory = sessionmaker(
+    # Create read-write session factory for PostgreSQL
+    async_rw_session_factory = sessionmaker(
         postgres_engine, class_=AsyncSession, expire_on_commit=False
+    )
+
+    # Create read-only session factory that enforces no writes at DB level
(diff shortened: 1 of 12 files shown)


agentex/tests/integration/fixtures/integration_client.py
```diff
@@ -229,11 +229,27 @@ async def isolated_repositories(isolated_test_schema):
...
+    # Create read-only session factory that enforces no writes at DB level
+    # Any INSERT, UPDATE, DELETE will fail with PostgreSQL error:
+    # "cannot execute X in a read-only transaction"
+    class ReadOnlyAsyncSession(AsyncSession):
+        """AsyncSession that sets PostgreSQL transaction to read-only mode."""
+
+        async def __aenter__(self):
+            session = await super().__aenter__()
+            # Set transaction to read-only mode - PostgreSQL will reject any writes
+            await session.execute(text("SET TRANSACTION READ ONLY"))
```
smoreinis, line 246: Ideally, this should guarantee that we won't get errors here during deployment when the read only session maker actually starts hitting the read replicas.

PR #118, opened 2025-12-24
perf: Add Cache-Control headers to GET endpoints

smoreinis:
## Summary
- Adds reusable `@cacheable` decorator for HTTP cache headers
- Applies cache headers to 9 GET endpoints across 4 route files
- Reduces server load by enabling client-side caching

## Changes
- **New `src/api/cache.py`**: Decorator supporting `max_age`, `private/public`, `stale-while-revalidate`
- **agents.py**: 60s cache for agent data (3 endpoints)
- **schedules.py**: 60s cache for schedule config (2 endpoints)
- **spans.py**: 300s cache for historical trace data (2 endpoints)
- **deployment_history.py**: 300s cache for historical deployments (2 endpoints)

## Usage
```python
@router.get("/items/{id}")
@cacheable(max_age=60)
async def get_item(id: str, response: Response) -> Item:
    ...
```

## Test plan
- [x] All existing tests pass (184/184)
- [x] Manual verification: Cache-Control header present

```bash
$ curl -i http://localhost:5003/spans 2>&1 | grep -i cache
cache-control: private, max-age=300
```

### DIFF
--- a/agentex/src/temporal/run_healthcheck_workflow.py
+++ b/agentex/src/temporal/run_healthcheck_workflow.py
@@ -7,6 +7,7 @@ from src.adapters.temporal.exceptions import (
 )
 from src.config.dependencies import (
     GlobalDependencies,
+    database_async_read_only_session_maker,
     database_async_read_write_engine,
     database_async_read_write_session_maker,
 )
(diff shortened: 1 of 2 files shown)


smoreinis:
hm.. agree about traces, I did think that deployment history would change relatively infrequently (with each deploy so like .. a few times a day at most) but thinking about it today I guess when it does change it makes more sense to display those immediately instead of waiting 5 minutes. will abandon this PR for now and revisit once there are better stats on which endpoints are actually getting the majority of the read traffic.

PR #121, opened 2025-12-24
Add proper readiness checks to health interceptor

smoreinis:
## Summary

- Separates liveness (`/healthz`) from readiness (`/readyz`, `/healthcheck`) probes
- Liveness probe returns 200 immediately (sub-millisecond) without checking dependencies
- Readiness probes check PostgreSQL, Redis, and MongoDB connectivity
- Returns 503 with detailed status when dependencies are unhealthy or not initialized
- Adds per-check timeout (2s) and overall timeout (5s) to prevent hung health checks

### Why this matters
Previously, all health check paths returned 200 regardless of dependency health. This meant Kubernetes would continue routing traffic to pods that couldn't actually serve requests. Now:
- `/healthz` (liveness): Fast response for Kubernetes to detect stuck processes
- `/readyz` and `/healthcheck` (readiness): Actual dependency validation before accepting traffic

### Response format (readiness)
```json
{
  "status": "ok",  // or "degraded"
  "checks": {
    "postgres": {"healthy": true},
    "redis": {"healthy": true},
    "mongodb": {"healthy": true}
  }
}
```

## Test plan

- [x] Unit tests verify liveness returns 200 without checking deps
- [x] Unit tests verify readiness returns 503 when deps unavailable
- [x] Unit tests verify readiness returns 200 when all deps healthy
- [x] All 188 unit tests pass


agentex/src/api/health_interceptor.py
```diff
@@ -37,37 +56,193 @@ def __init__(self, app: ASGIApp) -> None:
...
+        from sqlalchemy import text
+
+        try:
+            engine = deps.database_async_read_write_engine
+            if engine is None:
+                return {"healthy": False, "error": "Engine not initialized"}
+
+            async with asyncio.timeout(DEPENDENCY_CHECK_TIMEOUT):
+                async with engine.connect() as conn:
+                    await conn.execute(text("SELECT 1"))
```
RoxyFarhad, line 168: random q: no pings for PostgreSQL?

smoreinis: not sure what you mean by "pings" here? connecting to the engine feels similar (making sure the instance is reachable) and trying to run this basic query helps us confirm that the instance is actually handling requests as expected (as opposed to, say, being unresponsive)

RoxyFarhad: sorry, I meant is there no equivalent on a DB PING (like we do for mongo or redis) for postgresql? this is obviously good too

PR #146, opened 2026-02-08
feat: HTTP-proxy LangGraph checkpoint API

danielmillerp:
## What this does

This PR adds backend support for LangGraph checkpoint persistence — the mechanism LangGraph uses to save and restore agent state between messages (conversation history, channel values, pending writes, etc.).

### Why we need this

LangGraph agents need to persist their state (checkpoints) to a database. The built-in approach (`AsyncPostgresSaver`) has each agent connect directly to Postgres with its own connection pool. This doesn't scale — as we spin up more LangGraph agent pods, we'd hit connection limits quickly. This is the same problem we already solved for Temporal: instead of agents talking to the DB directly, they go through the backend API, which uses a shared connection pool.

### Why Postgres (not MongoDB)

Even though agent state currently lives in MongoDB, we chose Postgres for checkpoint storage. There have been some reliability concerns around MongoDB recently and there's a potential future migration to Postgres. Keeping new storage in Postgres is more future-forward. The checkpoint tables are independent and don't conflict with existing MongoDB state storage.

### How it works

**The pattern mirrors what we do with Temporal.** The agent doesn't know about the database — it talks to the backend API, and the backend handles the DB operations.

```
Agent (SDK HttpCheckpointSaver)  →  Backend API (/checkpoints/*)  →  Postgres
```

On the backend side, we:
1. Created **4 new Postgres tables** via ORM models + Alembic migration (`checkpoints`, `checkpoint_blobs`, `checkpoint_writes`, `checkpoint_migrations`) — these mirror the schema that LangGraph's own `AsyncPostgresSaver` uses
2. Built a **repository layer** that reimplements the same SQL operations from `AsyncPostgresSaver` using our SQLAlchemy patterns (composite primary keys, JSONB metadata, upserts via `ON CONFLICT`)
3. Exposed **5 POST endpoints** under `/checkpoints` (`get-tuple`, `put`, `put-writes`, `list`, `delete-thread`) — one for each method on LangGraph's `BaseCheckpointSaver`
4. Added **19 integration tests** for the repository layer, running against a real Postgres via testcontainers — covering round-trip storage, blob versioning, pending writes (upsert vs skip), metadata filtering, pagination, thread/namespace isolation, and deletion

Binary blob data (serialized Python objects) is base64-encoded for JSON transport. The actual serialization/deserialization stays in the SDK — the backend just stores and retrieves raw JSONB + bytes.

### Companion PRs
- **SDK (scale-agentex-python):** https://github.com/scaleapi/scale-agentex-python/pull/258 — replaces `AsyncPostgresSaver` with `HttpCheckpointSaver` that calls these endpoints
- **Agent code:** No changes needed — `create_checkpointer()` API is unchanged

## Test plan
- [x] 19 integration tests passing against real Postgres (testcontainers)
- [x] Manually tested end-to-end: backend + langgraph agent, sent messages, confirmed checkpoint stored and conversation history restored
- [x] Verified ruff lint + format passes

### DIFF
--- a/agentex/src/adapters/orm.py
+++ b/agentex/src/adapters/orm.py
@@ -5,6 +5,8 @@ from sqlalchemy import (
     DateTime,
     ForeignKey,
     Index,
+    Integer,
+    LargeBinary,
     String,
     Text,
     func,
@@ -212,4 +214,57 @@ class DeploymentHistoryORM(BaseORM):
             "ix_deployment_history_commit_hash",
             "commit_hash",
         ),
+    )
+
+
+# LangGraph checkpoint tables
+# These mirror the schema from langgraph.checkpoint.postgres so that
+# tables are created via Alembic migrations rather than at agent runtime.
+
+
+class CheckpointMigrationORM(BaseORM):
+    __tablename__ = "checkpoint_migrations"
+    v = Column(Integer, primary_key=True)
+
+
+class CheckpointORM(BaseORM):
+    __tablename__ = "checkpoints"
+    thread_id = Column(Text, nullable=False, primary_key=True)
+    checkpoint_ns = Column(Text, nullable=False, primary_key=True, server_default="")
+    checkpoint_id = Column(Text, nullable=False, primary_key=True)
+    parent_checkpoint_id = Column(Text, nullable=True)
+    type = Column(Text, nullable=True)
+    checkpoint = Column(JSONB, nullable=False)
+    metadata_ = Column("metadata", JSONB, nullable=False, server_default="{}")
+    __table_args__ = (
+        Index("checkpoints_thread_id_idx", "thread_id"),
+    )
+
+
+class CheckpointBlobORM(BaseORM):
+    __tablename__ = "checkpoint_blobs"
+    thread_id = Column(Text, nullable=False, primary_key=True)
+    checkpoint_ns = Column(Text, nullable=False, primary_key=True, server_default="")
+    channel = Column(Text, nullable=False, primary_key=True)
+    version = Column(Text, nullable=False, primary_key=True)
+    type = Column(Text, nullable=False)
+    blob = Column(LargeBinary, nullable=True)
+    __table_args__ = (
+        Index("checkpoint_blobs_thread_id_idx", "thread_id"),
+    )
+
+
+class CheckpointWriteORM(BaseORM):
+    __tablename__ = "checkpoint_writes"
+    thread_id = Column(Text, nullable=False, primary_key=True)
+    checkpoint_ns = Column(Text, nullable=False, primary_key=True, server_default="")
+    checkpoint_id = Column(Text, nullable=False, primary_key=True)
+    task_id = Column(Text, nullable=False, primary_key=True)
+    idx = Column(Integer, nullable=False, primary_key=True)
+    channel = Column(Text, nullable=False)
+    type = Column(Text, nullable=True)
+    blob = Column(LargeBinary, nullable=False)
+    task_path = Column(Text, nullable=False, server_default="")
+    __table_args__ = (
+        Index("checkpoint_writes_thread_id_idx", "thread_id"),
     )
--- a/agentex/src/api/routes/checkpoints.py
+++ b/agentex/src/api/routes/checkpoints.py
@@ -1,0 +1,206 @@
+import base64
+
+from fastapi import APIRouter, Response
+
+from src.api.schemas.checkpoints import (
+    BlobResponse,
+    CheckpointListItem,
+    CheckpointTupleResponse,
+    DeleteThreadRequest,
+    GetCheckpointTupleRequest,
+    ListCheckpointsRequest,
+    PutCheckpointRequest,
+    PutCheckpointResponse,
+    PutWritesRequest,
+    WriteResponse,
+)
+from src.api.schemas.authorization_types import (
+    AgentexResourceType,
+    AuthorizedOperationType,
+)
+from src.domain.use_cases.checkpoints_use_case import DCheckpointsUseCase
+from src.utils.authorization_shortcuts import DAuthorizedBodyId
+from src.utils.logging import make_logger
+
+logger = make_logger(__name__)
+
+router = APIRouter(prefix="/checkpoints", tags=["Checkpoints"])
+
+
+def _bytes_to_b64(data: bytes | None) -> str | None:
+    if data is None:
+        return None
+    return base64.b64encode(data).decode("ascii")
+
+
+def _b64_to_bytes(data: str | None) -> bytes | None:
+    if data is None:
+        return None
+    return base64.b64decode(data)
+
+
+@router.post(
+    "/get-tuple",
+    response_model=CheckpointTupleResponse | None,
+)
+async def get_checkpoint_tuple(
+    request: GetCheckpointTupleRequest,
+    checkpoints_use_case: DCheckpointsUseCase,
+    _authorized_task_id: DAuthorizedBodyId(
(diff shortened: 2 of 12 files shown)


agentex/src/adapters/orm.py
```diff
@@ -213,3 +215,56 @@ class DeploymentHistoryORM(BaseORM):
...
         ),
     )
+
+
+# LangGraph checkpoint tables
+# These mirror the schema from langgraph.checkpoint.postgres so that
+# tables are created via Alembic migrations rather than at agent runtime.
+
+
+class CheckpointMigrationORM(BaseORM):
```
RoxyFarhad, line 225: what is this?

agentex/src/adapters/orm.py
```diff
@@ -213,3 +215,56 @@ class DeploymentHistoryORM(BaseORM):
...
+    type = Column(Text, nullable=False)
+    blob = Column(LargeBinary, nullable=True)
+    __table_args__ = (
+        Index("checkpoint_blobs_thread_id_idx", "thread_id"),
+    )
+
+
+class CheckpointWriteORM(BaseORM):
+    __tablename__ = "checkpoint_writes"
+    thread_id = Column(Text, nullable=False, primary_key=True)
```
RoxyFarhad, line 259: can these be references to like the "parent" table so there is some synchronization between the tables? not sure if this is the right one but there seems to be a lot of duplicate fields here. It would be better if we have all of those in a single table, and then the tables that have other fields just referenced them/ 

That would better manage data and help with atomicity

agentex/src/api/routes/checkpoints.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def _b64_to_bytes(data: str | None) -> bytes | None:
+    if data is None:
+        return None
+    return base64.b64decode(data)
+
+
+@router.post(
+    "/get-tuple",
```
RoxyFarhad, line 43: how did you decide these routes for the function? are these langgraph patterns? 

they just aren't very clear what they are doing (like put-write also doesn't make sense to me if its not directly tied to a langgraph thing)

agentex/src/domain/repositories/checkpoint_repository.py
```diff
@@ -0,0 +1,308 @@
...
+        before_checkpoint_id: str | None = None,
+        filter_metadata: dict[str, Any] | None = None,
+        limit: int | None = None,
+    ) -> list[dict[str, Any]]:
+        """List checkpoints matching criteria, ordered newest first."""
+        async with (
+            self.async_ro_session_maker() as session,
+            async_sql_exception_handler(),
+        ):
+            query = select(CheckpointORM)
```
RoxyFarhad: if none of the filters are set does this load the entire table in memory? if so, that could be bad for other requests on server / latency of other request paths hitting the DB

agentex/src/domain/repositories/checkpoint_repository.py
```diff
@@ -0,0 +1,308 @@
...
+                        "parent_checkpoint_id": cp.parent_checkpoint_id,
+                        "checkpoint": cp.checkpoint,
+                        "metadata": cp.metadata_,
+                    }
+                )
+            return checkpoints
+
+    async def delete_thread(self, thread_id: str) -> None:
+        """Delete all checkpoint data for a thread."""
+        async with (
```
RoxyFarhad, line 287: see comment about changing the way we do do DB tables because this would be much cleaner with foreign keys, references etc

agentex/src/adapters/orm.py
danielmillerp: this is all part of Langgraph checkpointing. See: https://docs.langchain.com/oss/python/langgraph/persistence, https://pypi.org/project/langgraph-checkpoint-postgres/

agentex/src/api/routes/checkpoints.py
danielmillerp: this is implementing their base class

RoxyFarhad: can you maybe say that and link to something from langchain here so its clear

agentex/src/api/routes/checkpoints.py
```diff
@@ -0,0 +1,186 @@
...
+    if data is None:
+        return None
+    return base64.b64encode(data).decode("ascii")
+
+
+def _b64_to_bytes(data: str | None) -> bytes | None:
+    if data is None:
+        return None
+    return base64.b64decode(data)
+
```
RoxyFarhad, line 40: comment to validate permissions models here (and that if you do not have access to an agent, you should not be able to see their checkpoints)

danielmillerp: just checked this api is consistent with x-agent-api-key header which is how we make sure an agent can call agentex server

as far as like permissions per tasks or threads for langgraph we dont have anything in place , not just for langgraph but generally, Agent A can get Agent B's task in Agentex

agentex/src/api/routes/checkpoints.py
danielmillerp: sure!

agentex/src/api/routes/checkpoints.py
danielmillerp: All 5 checkpoint endpoints now have DAuthorizedBodyId authorization checks using field_name="thread_id", which    
  works because thread_id = task.id (set in acp.py line 46: thread_id = params.task.id).                                  
                                                                                                                          
  The authorization checks:                                                                                               
  - /get-tuple and /list → read on the task                                                                               
  - /put and /put-writes → execute on the task (matching the messages pattern for write operations)                  
  - /delete-thread → delete on the task                                                                                   
                                                                                                                        
  For agent requests (via x-agent-api-key), these checks still bypass (same as all other endpoints). But for client       
  requests (via the auth gateway), the authorization service will now verify the caller has permission on that specific   
  task before allowing checkpoint access. This is consistent with how messages and states protect their endpoints.

agentex/src/domain/repositories/checkpoint_repository.py
danielmillerp: adding a limit!

danielmillerp: and made thread id non-optional

agentex/src/domain/repositories/checkpoint_repository.py
danielmillerp: unfort cannot change constructs this is langgraph defined

agentex/src/adapters/orm.py
danielmillerp: langgraph constructs same as other comment not really up to us

agentex/database/migrations/alembic/versions/2026_02_07_0000_add_langgraph_checkpoint_tables_a1b2c3d4e5f6.py
```diff
@@ -0,0 +1,89 @@
+"""add_langgraph_checkpoint_tables
+
+Revision ID: a1b2c3d4e5f6
```
smoreinis: can we do a real revision ID here (like, by using make migration)?
running make migration will also update the migration history as well which didn't happen here

danielmillerp: for sure will do!

danielmillerp: done!

PR #270, opened 2026-06-03
feat(authz): register agentex resources and grant agent/task ownership

asherfink:
## Related work

Parent ticket: AGX1-242 - AgentEx authorization dual-write.

This change is part of a 4-PR stack across 3 repos.

| Repo | PR | Purpose |
|---|---|---|
| scaleapi/scaleapi | [scaleapi/scaleapi#146335](https://github.com/scaleapi/scaleapi/pull/146335) | merged flag registry update for the shared AgentEx resources rollout flags |
| scaleapi/agentex | [scaleapi/agentex#364](https://github.com/scaleapi/agentex/pull/364) | agentex-auth routes each auth verb to one backend target |
| **scaleapi/scale-agentex** | **this PR** | **registers agentex resources in the authorization graph and grants agent/task ownership** |
| scaleapi/scale-agentex | [scaleapi/scale-agentex#271](https://github.com/scaleapi/scale-agentex/pull/271) | denied direct agent access collapses to 404 |

**Merge/deploy order:** [scaleapi/scaleapi#146335](https://github.com/scaleapi/scaleapi/pull/146335) is merged; deploy [scaleapi/agentex#364](https://github.com/scaleapi/agentex/pull/364) before this PR, then merge/deploy [scaleapi/scale-agentex#271](https://github.com/scaleapi/scale-agentex/pull/271). Rollout per account is: enable `fgac-agentex-resources-dual-write`, backfill existing resources into Spark, then enable `fgac-agentex-resources` reads.

## What

Registers agentex resources in the authorization graph on create and removes them on delete, separating resource lifecycle from explicit ownership:

- **Agents and tasks** register in the authorization graph on create and record an explicit owner; on delete they deregister and the owner record is revoked.
- **Schedules and agent API keys** register under their parent agent on create and deregister on delete. Permissions cascade from the parent agent, so no separate owner record is written.

Registration runs before the resource is persisted, so an authorization failure fails closed with no orphaned row. If the persist (or the Temporal create, for schedules) later fails, a compensating deregister runs.

## Why

Giving each verb a single backend keeps the existing ownership writes (`grant`/`revoke`) unchanged and confines the new behavior to the new lifecycle interface (`register`/`deregister`). Explicit ownership is recorded only for agents and tasks; schedules and agent API keys inherit access from their parent agent, so they need no owner record of their own. Issuing the two write kinds independently keeps both backends current, so reads can cut over per account with a clean rollback.

## Testing

- Unit and integration tests covering the agent, task, schedule, and api_key authorization writes (register/deregister ordering, fail-closed-before-persist, and compensating cleanup).
- `ruff check` / `ruff format`.

<h3>Greptile Summary</h3>

This PR adds `register_resource` / `deregister_resource` lifecycle calls to agents, tasks, schedules, and agent API keys so that each resource is tracked in the authorization graph from the moment it is created. `register_resource` fires before the Postgres (or Temporal) write so an auth failure fails closed with no orphaned row; a compensating `deregister_resource` cleans up if the downstream write later fails. Ownership writes (`grant` / `revoke`) remain at the route layer, keeping both backends current during the dual-write rollout.

- **AgentsUseCase** gains `_register_in_auth` (called only on genuine first creates, not on updates/idempotent re-registers) and `_safe_deregister` (best-effort, called after soft-delete and as compensation for race conditions or persist failures).
- **ScheduleService** and **AgentAPIKeysUseCase** already had this pattern; their docstrings and comments are updated to reflect the unified terminology.
- `principal_context` is removed from agent schemas because the authorization service now derives the principal from its own injected context.

Safe to merge; register/deregister ordering is correct across all resource types and all failure paths are covered with compensating actions or best-effort swallowing.

The core invariant — register before persist so an auth failure aborts the create cleanly — holds in both AgentsUseCase create paths. Compensation on DuplicateItemError and generic Exception correctly deregisters only when a register call was actually made. The delete path deregisters after the soft-delete/Postgres delete, consistent with existing task and schedule patterns. Tests cover the full lifecycle including race conditions and the no-creator skip path.

</details>

| Filename | Overview |
|----------|----------|
| agentex/src/domain/use_cases/agents_use_case.py | Adds DAuthorizationService dependency, _register_in_auth (skips when no principal, fail-closed before persist), and _safe_deregister (swallows errors). Both genuine-create paths call _register_in_auth with correct compensation on DuplicateItemError and generic Exception; update paths are correctly excluded. |
| agentex/src/api/routes/agents.py | Removes per-resource post-fetch auth check from get_agent_by_name (pre-check dependency remains); removes principal_context forwarding; switches register_build from register_resource to grant at route level. |
| agentex/tests/integration/use_cases/test_agent_authz_dual_write.py | New integration test covering all key scenarios: register before persist, fail-closed on register failure, compensation on persist failure, duplicate-race compensation, update path exclusion, no-creator skip, and delete deregistration. |
| agentex/tests/unit/api/test_tasks_authz.py | Adds TestTaskDeleteAuthWrites covering delete_task and delete_task_by_name routes, verifying revoke is called with the correctly resolved task ID. |
| agentex/src/api/schemas/agents.py | Removes principal_context field from RegisterAgentRequest and RegisterBuildRequest; reflected in openapi.yaml. |

</details>

<sub>Reviews (9): Last reviewed commit: ["feat(authz): register agentex resources ..."](https://github.com/scaleapi/scale-agentex/commit/73a5056fbc79cf60a133b7bf079e5c291249e2ee) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35397368)</sub>

### DIFF
--- a/agentex/src/api/routes/agents.py
+++ b/agentex/src/api/routes/agents.py
@@ -76,16 +76,9 @@ async def get_agent_by_name(
         AgentexResourceType.agent, AuthorizedOperationType.read
     ),
     agents_use_case: DAgentsUseCase,
-    authorization: DAuthorizationService,
 ):
     """Get an agent by its unique name."""
     agent_entity = await agents_use_case.get(name=agent_name)
-
-    await authorization.check(
-        resource=AgentexResource.agent(agent_entity.id),
-        operation=AuthorizedOperationType.read,
-    )
-
     return Agent.model_validate(agent_entity)
 
 
@@ -183,7 +176,6 @@ async def register_agent(
     await authorization_service.check(
         AgentexResource.agent("*"),
         AuthorizedOperationType.create,
-        principal_context=request.principal_context,
     )
     logger.info(f"Registering agent: {request}")
     try:
@@ -198,7 +190,6 @@ async def register_agent(
         )
         await authorization_service.grant(
             AgentexResource.agent(agent_entity.id),
-            principal_context=request.principal_context,
         )
         response_fields = agent_entity.model_dump()
         existing_key = await api_keys_use_case.get_internal_api_key_by_agent_id(
@@ -236,11 +227,10 @@ async def register_build(
     agents_use_case: DAgentsUseCase,
     authorization_service: DAuthorizationService,
 ) -> Agent:
-    """Create a build-only agent row and register its authz resource."""
+    """Create a build-only agent row after create authorization."""
     await authorization_service.check(
         AgentexResource.agent("*"),
         AuthorizedOperationType.create,
-        principal_context=request.principal_context,
     )
     logger.info(f"Registering build for agent: {request.name}")
     try:
@@ -252,9 +242,8 @@ async def register_build(
         )
     except ValueError as e:
         raise HTTPException(status_code=400, detail=str(e)) from e
-    await authorization_service.register_resource(
+    await authorization_service.grant(
         AgentexResource.agent(agent_entity.id),
-        principal_context=request.principal_context,
     )
     return Agent.model_validate(agent_entity)
 
--- a/agentex/openapi.yaml
+++ b/agentex/openapi.yaml
@@ -5221,12 +5221,6 @@ components:
         acp_type:
           $ref: '#/components/schemas/ACPType'
           description: The type of ACP to use for the agent.
-        principal_context:
-          anyOf:
-          - {}
-          - type: 'null'
-          title: Principal Context
-          description: Principal used for authorization
         registration_metadata:
           anyOf:
           - additionalProperties: true
@@ -5336,12 +5330,6 @@ components:
           type: string
           title: Description
           description: The description of the agent.
-        principal_context:
-          anyOf:
-          - {}
-          - type: 'null'
-          title: Principal Context
-          description: Principal used for authorization
         registration_metadata:
           anyOf:
           - additionalProperties: true
--- a/agentex/src/api/schemas/agents.py
+++ b/agentex/src/api/schemas/agents.py
@@ -88,9 +88,6 @@ class RegisterAgentRequest(BaseModel):
         description="Optional agent ID if the agent already exists and needs to be updated.",
     )
     acp_type: ACPType = Field(..., description="The type of ACP to use for the agent.")
-    principal_context: Any | None = Field(
-        default=None, description="Principal used for authorization"
-    )
     registration_metadata: dict[str, Any] | None = Field(
         default=None,
         description="The metadata for the agent's registration.",
@@ -120,9 +117,6 @@ class RegisterBuildRequest(BaseModel):
         ..., pattern=r"^[a-z0-9-]+$", description="The unique name of the agent."
     )
     description: str = Field(..., description="The description of the agent.")
-    principal_context: Any | None = Field(
-        default=None, description="Principal used for authorization"
-    )
     registration_metadata: dict[str, Any] | None = Field(
         default=None,
         description="The metadata for the agent's build registration.",
--- a/agentex/src/domain/services/schedule_service.py
+++ b/agentex/src/domain/services/schedule_service.py
@@ -84,13 +84,12 @@ class ScheduleService:
             else None
         )
 
-        # Schedules have no Postgres row — Temporal is the store and the auth
+        # Schedules have no Postgres row: Temporal is the store and the auth
         # selector is the schedule id ({agent_id}--{schedule_name}). Register
-        # the auth tuple (with the parent_agent edge) BEFORE the Temporal write
-        # (fail-closed), and compensate if the Temporal create fails so we never
-        # leave an orphan tuple. The read-back below is intentionally OUTSIDE
-        # the compensation scope: a describe failure must not deregister a
-        # schedule that was actually created.
+        # before the Temporal write so an auth failure fails closed. If the
+        # Temporal create fails after registration, compensate with a deregister.
+        # The read-back below is intentionally outside the compensation scope
+        # because the schedule was already created.
         registered = await self._register_schedule_in_auth(
             schedule_id=schedule_id, agent_id=agent.id
         )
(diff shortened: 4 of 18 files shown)


agentex/src/api/routes/agents.py
```diff
@@ -76,16 +76,9 @@ async def get_agent_by_name(
         AgentexResourceType.agent, AuthorizedOperationType.read
     ),
     agents_use_case: DAgentsUseCase,
-    authorization: DAuthorizationService,
 ):
     """Get an agent by its unique name."""
     agent_entity = await agents_use_case.get(name=agent_name)
-
-    await authorization.check(
```
harvhan, line 84: I think we should keep the authorization.check/revoke/grant. Otherwise, how can we maintain backward compatibility?

1, removing authorization.check will remove authz altogether.
2, removing grant/revoke will break the old authz even if the user opts out of the new FGAC

Even if a user turns FGAC on, we may still want to keep grant/revoke for permission data dual-write, so that the old authz still works in case the user turns FGAC off.

asherfink: went and traced this through `agentex-auth` to make sure we're solid, and I don't think we need to keep grant/revoke here

**tl;dr:** on the legacy (SGP) backend, `register_resource` *is* `grant` and `deregister_resource` *is* `revoke`. so dropping the explicit grant/revoke and going through register/deregister is behavior-preserving for old authz, not a regression.

> **1. removing authorization.check removes authz altogether**

it doesn't. the only `check` I removed is the explicit one in `get_agent_by_name`, but that route still has `agent_name: DAuthorizedName(agent, read)` in its signature, and that dependency calls `authorization.check(agent, read)` itself before the handler body runs ([authorization_shortcuts.py](https://github.com/scaleapi/scale-agentex/blob/f4e7e86300549b214018e62ac09c984702e57a60/agentex/src/utils/authorization_shortcuts.py#L195)). so the line I deleted was a redundant second check of the same tuple. read authz is still enforced. on create I kept the `check(agent("*"), create)` too, just dropped the unused `principal_context=` kwarg.

> **2 + 3. removing grant/revoke breaks old authz when FGAC is off**

two things:

- `agentex-auth` routes every call (grant/revoke/check/register/deregister) to exactly *one* provider per account based on `FGAC_AGENTEX_AUTH_SPARK` ([`_resolve_provider`](https://github.com/scaleapi/agentex/blob/1ddd7593122ebf4317c123792248550ed8fad317/agentex-auth/src/domain/services/authorization_service.py#L56)). flag off → legacy SGP, flag on → Spark. there's no write-to-both happening, it's a routing switch.
- on the SGP gateway, [`register_resource` delegates straight to `grant`](https://github.com/scaleapi/agentex/blob/1ddd7593122ebf4317c123792248550ed8fad317/agentex-auth/src/adapters/authorization/adapter_sgp_authorization.py#L155) and `deregister_resource` delegates to `revoke`

so for an FGAC-off account, `register_resource(agent)` → SGP → `grant(owner)`, which is exactly what the old route did. `deregister_resource` → `revoke`. legacy authz is fully preserved. for FGAC-on it does the proper Spark resource+owner write (tenant scoping comes from account_id automatically). either way it's correct, and keeping grant/revoke on top would just be a redundant duplicate write.

asherfink:
didn't mean to delete the greptile comments from^^, claude did that as they were from the old impl

agentex/src/api/routes/agents.py
```diff
@@ -76,16 +76,9 @@ async def get_agent_by_name(
...
-        resource=AgentexResource.agent(agent_entity.id),
-        operation=AuthorizedOperationType.read,
-    )
-
```
rpatel-scale, line 88: Why are we removing this? Does it now happen at a higher layer?

agentex/src/api/routes/agents.py
```diff
@@ -252,9 +242,8 @@ async def register_build(
         )
     except ValueError as e:
         raise HTTPException(status_code=400, detail=str(e)) from e
-    await authorization_service.register_resource(
+    await authorization_service.grant(
```
rpatel-scale, line 245: So we do a register resource in agentex/src/domain/use_cases/agents_use_case.py and then a grant out here. Do we need to do both? Can we keep the authz checks/grants at the router layer ideally instead of nested in the business logic? Thoughts?

asherfink: we need both, they're the two backends of the dual-write, not redundant. `register_resource` records the agent in the new authz graph; `grant` records the legacy owner edge.

agentex/src/api/routes/agents.py
asherfink: it's moved up into the parameter dependency, so the body call was a duplicate. `agent_name` is typed `DAuthorizedName(AgentexResourceType.agent, read)`, and that dependency (`_ensure_authorized_name` in `src/utils/authorization_shortcuts.py`) resolves the agent by name + runs `authorization.check(agent(resolved_id), read)` before the handler body executes.

get_agent_by_id directly above already works this way

agentex/src/api/routes/agents.py
rpatel-scale: sure makes sense but still can we move it up from the business layer? or move this call down. Ideally all authz checks and writes happen at the highest layer possible IMO

asherfink: agree on the principle for the *checks*. gating belongs at the edge and it already is there: the `check` runs in the route and in the `DAuthorizedId` / `DAuthorizedName` param dependencies (which resolve *before* the handler body executes), so access control happens at the highest layer.

`register_resource` / `deregister_resource` are not authz gating. they're resource-graph writes that have to be transactionally consistent with the DB persist:

- run *before* the row is persisted (fail-closed → no orphaned row),
- compensate with a `deregister` if the persist loses a duplicate-create race or fails,
- fire only on a genuine create (the get-or-create logic in the use case is what knows create-vs-update; re-registering on an update would rewrite ownership).

The route only gets control *after* `register_agent` returns (post-persist), and the agent id is generated inside it. so it can't do register-before-persist or know create-vs-update without exposing the transaction internals upward. Moving it to the router would pull persistence/race logic into the route, which is worse

rpatel-scale (approved):

rpatel-scale (approved):

agentex/src/api/routes/agents.py
harvhan: this was duplicative even before this PR.

asherfink:
rebased onto latest `main`

PR #278, opened 2026-06-04
fix(states): update states by state ID only

deepthi-rao-scale:
## Summary
- make state updates identify the row only by the path `state_id`
- keep legacy `task_id` and `agent_id` update payloads accepted but ignored
- rely on `DAuthorizedId(TaskChildResourceType.state, update)` to resolve `state_id` to the stored parent `task_id` and authorize against that task before the handler runs
- update only replaces the persisted state's `.state` field; stored `task_id` and `agent_id` are preserved by the read-modify-write flow
- return 404 when the state lookup fails, and collapse parent-task auth denial to 404 through the existing authorization helper

## Verification
- `uv run --group test pytest tests/unit/api/test_states_schema.py tests/unit/use_cases/test_states_use_case.py -q`

Note: the focused local state API integration test was attempted earlier, but Docker/Testcontainers could not reach the local Docker socket before executing the test body. GitHub Actions integration shards passed.

<h3>Greptile Summary</h3>

This PR simplifies the state-update path by removing `task_id` / `agent_id` from `UpdateStateRequest` and relying solely on the `state_id` path parameter. Authorization is handled upstream by `DAuthorizedId(TaskChildResourceType.state, update)`, which looks up the parent task and checks the caller's permissions before the handler runs.

- **Schema cleanup**: `UpdateStateRequest` drops the two previously-required parent-identifier fields; extra fields sent by legacy clients are silently ignored (confirmed by the new unit test).
- **Use-case hardening**: `StatesUseCase.update` now raises `ItemDoesNotExist` (→ 404) instead of silently returning `None` when the state is not found, providing defensive protection for race conditions between the auth lookup and the handler.
- **Test coverage**: Two new unit-test modules cover the schema ignore-behaviour and the use-case not-found path; the integration test is updated to reflect that body-level parent identifiers are now ignored rather than enforced.

Safe to merge — the change correctly narrows the update contract to state_id-only routing, with authorization handled entirely by the existing DAuthorizedId dependency.

All changed files are consistent with each other: schema, route, use case, OpenAPI spec, and tests all agree on the new contract. The defensive ItemDoesNotExist guard in the use case correctly handles race conditions. Legacy clients sending task_id/agent_id in the request body continue to work without errors. New unit tests verify both the happy path and the not-found path.

</details>

| Filename | Overview |
|----------|----------|
| agentex/src/domain/use_cases/states_use_case.py | Removes task_id parameter from update(); adds ItemDoesNotExist guard for the not-found case. Logic is clean and tests cover both paths. |
| agentex/src/api/routes/states.py | Drops task_id forwarding to the use case; auth is fully delegated to DAuthorizedId. No issues. |
| agentex/src/api/schemas/states.py | task_id and agent_id removed from UpdateStateRequest; BaseModel ignores unknown extra fields, preserving backwards compatibility for legacy clients. |
| agentex/openapi.yaml | UpdateStateRequest schema updated to remove task_id/agent_id from both properties and required array; consistent with schema change. |
| agentex/tests/unit/use_cases/test_states_use_case.py | New unit tests cover update-by-id happy path and ItemDoesNotExist raise; well-structured with AsyncMock and pytest-asyncio. |
| agentex/tests/unit/api/test_states_schema.py | Validates that extra legacy fields in the request body are silently dropped by UpdateStateRequest. |
| agentex/tests/integration/api/states/test_states_api.py | Removes the stale 'wrong task_id is a no-op' test and replaces it with a test confirming legacy body fields are accepted but ignored. |

</details>

```mermaid
sequenceDiagram
    participant Client
    participant DAuthorizedId
    participant _get_parent_task_id
    participant StateRepo
    participant AuthService
    participant StatesUseCase

    Client->>DAuthorizedId: "PUT /states/{state_id}"
    DAuthorizedId->>_get_parent_task_id: resolve state_id
    _get_parent_task_id->>StateRepo: "get(id=state_id)"
    StateRepo-->>_get_parent_task_id: StateEntity
    _get_parent_task_id-->>DAuthorizedId: task_id
    DAuthorizedId->>AuthService: check_task_or_collapse_to_404
    AuthService-->>DAuthorizedId: authorized
    DAuthorizedId-->>Client: state_id resolved
    Client->>StatesUseCase: "update(id=state_id, state=new_state)"
    StatesUseCase->>StateRepo: "get(id=state_id)"
    StateRepo-->>StatesUseCase: StateEntity
    StatesUseCase->>StateRepo: update(task_state)
    StateRepo-->>StatesUseCase: updated StateEntity
    StatesUseCase-->>Client: 200 State
```
</details>

<sub>Reviews (2): Last reviewed commit: ["chore(states): update OpenAPI spec"](https://github.com/scaleapi/scale-agentex/commit/d4566de69ac2194ce0346184f43721af767517ce) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35740604)</sub>

### DIFF
--- a/agentex/src/domain/use_cases/states_use_case.py
+++ b/agentex/src/domain/use_cases/states_use_case.py
@@ -2,6 +2,7 @@ from typing import Annotated, Any
 
 from fastapi import Depends
 
+from src.adapters.crud_store.exceptions import ItemDoesNotExist
 from src.domain.entities.states import StateEntity
 from src.domain.repositories.task_state_repository import DTaskStateRepository
 from src.utils.logging import make_logger
@@ -58,13 +59,14 @@ class StatesUseCase:
             order_direction=order_direction,
         )
 
-    async def update(self, id: str, task_id: str, state: dict[str, Any]) -> StateEntity:
+    async def update(self, id: str, state: dict[str, Any]) -> StateEntity:
         task_state = await self.task_state_repository.get(id=id)
-        if task_state and task_state.task_id == task_id:
-            # Update the state field but preserve other fields
-            task_state.state = state
-            return await self.task_state_repository.update(task_state)
-        return task_state
+        if task_state is None:
+            raise ItemDoesNotExist(f"State {id} not found")
+
+        # Update the state field but preserve other fields.
+        task_state.state = state
+        return await self.task_state_repository.update(task_state)
 
     async def delete(self, id: str) -> None:
         return await self.task_state_repository.delete(id=id)
--- a/agentex/tests/integration/api/states/test_states_api.py
+++ b/agentex/tests/integration/api/states/test_states_api.py
@@ -190,31 +190,14 @@ class TestStatesAPIIntegration:
         assert created_state["agent_id"] == state_data["agent_id"]
         assert created_state["state"] == state_value
 
-        # API-first validation: UPDATE rejects parent identifiers in the body
-        update_response = await isolated_client.put(
-            f"/states/{state_id}",
-            json={
-                "state": {},
-                "task_id": "some-other-task-id",
-                "agent_id": test_agent.id,
-            },
-        )
-        assert update_response.status_code == 422
-
-        unchanged_response = await isolated_client.get(f"/states/{state_id}")
-        assert unchanged_response.status_code == 200
-        unchanged_state = unchanged_response.json()
-        assert unchanged_state["id"] == state_id
-        assert unchanged_state["task_id"] == state_data["task_id"]
-        assert unchanged_state["agent_id"] == state_data["agent_id"]
-        assert unchanged_state["state"] == state_value
-
-        # API-first validation: UPDATE the created state
+        # API-first validation: UPDATE accepts legacy parent identifiers in the body
         state_value_updated = {"test": "updated"}
         update_response = await isolated_client.put(
             f"/states/{state_id}",
             json={
                 "state": state_value_updated,
+                "task_id": "some-other-task-id",
+                "agent_id": test_agent.id,
             },
         )
         assert update_response.status_code == 200
--- a/agentex/tests/integration/api/states/test_states_api.py
+++ b/agentex/tests/integration/api/states/test_states_api.py
@@ -190,7 +190,7 @@ class TestStatesAPIIntegration:
         assert created_state["agent_id"] == state_data["agent_id"]
         assert created_state["state"] == state_value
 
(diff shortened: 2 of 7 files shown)


agentex/src/domain/use_cases/states_use_case.py
```diff
@@ -58,13 +59,14 @@ async def list(
             order_direction=order_direction,
         )
 
-    async def update(self, id: str, task_id: str, state: dict[str, Any]) -> StateEntity:
+    async def update(self, id: str, state: dict[str, Any]) -> StateEntity:
         task_state = await self.task_state_repository.get(id=id)
```
deepthi-rao-scale, line 63: do we need this check if the auth check already checked for existence of task id?

agentex/tests/integration/api/states/test_states_api.py
```diff
@@ -190,7 +190,7 @@ async def test_create_and_retrieve_state_consistency(
         assert created_state["agent_id"] == state_data["agent_id"]
         assert created_state["state"] == state_value
 
-        # API-first validation: UPDATE with wrong task_id does nothing
+        # API-first validation: UPDATE rejects parent identifiers in the body
```
deepthi-rao-scale: if we do this it will cause a regression of the existing calls, unintended behavior.

agentex/tests/unit/api/test_states_schema.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+from pydantic import ValidationError
+from src.api.schemas.states import UpdateStateRequest
+
+
+@pytest.mark.unit
+def test_update_state_request_accepts_state_only():
```
deepthi-rao-scale: why do we need these tests.

deepthi-rao-scale:
@greptile

agentex/src/domain/use_cases/states_use_case.py
```diff
@@ -58,13 +59,14 @@ async def list(
...
-        if task_state and task_state.task_id == task_id:
-            # Update the state field but preserve other fields
-            task_state.state = state
-            return await self.task_state_repository.update(task_state)
```
declan-scale, line 66: @smoreinis any reason not having the task_id would introduce a regression?

smoreinis: we should be fine here, thanks for double checking

declan-scale (approved):

PR #335, opened 2026-06-23
feat(schedules): agent run schedules (v1)

jromualdez-scale:
## Summary

Adds per-agent **run schedules**: recurring schedules that fire a task and deliver a configured initial input on a cron/interval cadence. **Replaces the prior `schedules` implementation** (a bare-workflow scheduler) on the same API path.

Each schedule is a Postgres row (the source of truth) plus a Temporal Schedule that acts purely as the recurring clock (it carries only the row id). On each fire, a thin, deterministic workflow runs a single activity that creates a task and delivers the initial input via the same path as a manual run — `message/send` for sync agents, `event/send` for agentic agents — attributed to the schedule's stored creator principal.

## Feature flag

The API is gated behind `ENABLE_AGENT_RUN_SCHEDULES` (matches the existing `ENABLE_HEALTH_CHECK_WORKFLOW` pattern), **disabled by default in every environment** — when off, the routes are not registered at all. Enable per-environment when ready to test (e.g. locally `ENABLE_AGENT_RUN_SCHEDULES=true ./dev.sh`). The OpenAPI spec/SDK document the endpoints regardless of the runtime default.

## Removed / breaking changes

This PR **deletes the previous `schedules` feature** (routes, schemas, service, use case, and its tests). The old endpoint scheduled a raw Temporal workflow and stored nothing in Postgres; the new one schedules an agent *run* and is Postgres-backed. Because the API path `/agents/{agent_id}/schedules` is reused with new semantics, this is **breaking for existing consumers of the old endpoint**:

- `POST /agents/{agent_id}/schedules` — request/response schema changed (schedules an agent run, not a bare workflow)
- `POST …/{name}/unpause` → **renamed** to `…/{name}/resume`
- Path param `{schedule_name}` → `{name}` (cosmetic)
- Adds the new `agent_run_schedules` table (the old scheduler was Temporal-only)

(`…/{name}/trigger` is preserved — see below.)

## Endpoints

`/agents/{agent_id}/schedules`:
- `POST` — create
- `GET` — list (served from Postgres; no per-row Temporal call)
- `GET /{name}` — get (includes live Temporal state: next/last fire, action count)
- `PATCH /{name}` — partial update (cadence, window, input, params, paused; cron/interval stay mutually exclusive)
- `POST /{name}/pause` · `POST /{name}/resume`
- `POST /{name}/trigger` — immediate out-of-band run
- `DELETE /{name}`

## Implementation notes

- `ScheduledAgentRunWorkflow` (thin/deterministic) + `launch_scheduled_agent_run` activity (all side effects live in the activity).
- Deterministic per-fire task name makes `task/create` idempotent on activity retry; a delivered marker guards against duplicate input delivery.
- Fire-time authorization re-check under the stored creator principal — a revoked creator stops firing cleanly.
- Scheduled tasks get a `task_metadata.display_name` (`Scheduled Message: <name> · <fire time>`), stamped with the nominal fire time (stable across retries) so they render with a label instead of "Unnamed task".
- `delete`/`pause`/`resume`/`update` tolerate a missing Temporal schedule so a partial failure can't strand an un-cleanable row.
- New `agent_run_schedules` table migration (new-table create; schema-only, non-blocking).

## Testing

- 30 unit tests (service, activity, use case, env flag) pass, covering create/list/get/update/pause/resume/trigger/delete, idempotency, validation, and flag parsing.
- Verified end-to-end locally (flag on): both delivery paths (sync `message/send` and agentic `event/send`), plus pause/resume/update/trigger/delete reflected consistently in Postgres and Temporal.
- Verified on a dev cluster (branch image, flag on): create → Temporal schedule → worker fires on schedule → `message/send` delivered, with the row persisted and the creator principal captured from real auth.

## Deployment dependency (authz provider)

Dev verification surfaced this: on a cluster using the SGP authz provider (`AUTH_PROVIDER=sgp`), **the provider must learn the new `schedule` resource type before this is usable there.** Today its `/v1/authz/check` returns **422** for a `schedule` resource, so:

- **Create works** (it gates on `agent.update`, and `register` of the `schedule` resource is tolerated).
- **Every op gated on the schedule resource — `GET /{name}`, `pause`, `resume`, `trigger`, `PATCH`, `DELETE` — returns 422** until the provider handles `check`/`grant`/`revoke`/`register`/`deregister`/`search` for `schedule` (mirroring `agent`/`task`/`api_key`).

This is provider-side work (the `schedule` type is already part of the documented auth-provider contract); it should land alongside this feature's rollout. Environments with authz disabled or a permissive provider are unaffected.

<h3>Greptile Summary</h3>

This PR replaces the previous bare-workflow scheduler with a fully Postgres-backed agent run scheduling system. Each schedule is a Postgres row (source of truth) paired with a Temporal Schedule acting as the recurring clock; on each fire, a thin deterministic workflow runs a single activity that creates a fresh task and delivers the configured initial input under the stored creator principal.

- Adds `agent_run_schedules` table, ORM, repository, service, use case, and full CRUD API (`POST`/`GET`/`PATCH`/`DELETE`/`pause`/`resume`/`trigger`) gated behind `ENABLE_AGENT_RUN_SCHEDULES`.
- Implements idempotent fire-and-deliver via deterministic task names and a `scheduled_input_delivered` marker; all previously flagged issues (partial-delete strand, list-path sequential Temporal RPCs, fire-time stamp drift, manual-trigger paused bypass, auth-filter/limit ordering, mutually-exclusive cadence validation) have been addressed in the current code.
- The Temporal worker unconditionally registers the new workflow and activities so that in-flight scheduled executions continue to be handled even when the API flag is disabled.

Safe to merge; the core fire-and-deliver path is well-guarded, all previously flagged issues are addressed, and the feature is off by default.

All six issues called out in prior review threads have been resolved in the current code. The new activity is idempotent (deterministic task name + delivered marker), the rollback logic on failed creates is correct, and the migration is non-blocking. Remaining observations are non-blocking design trade-offs with documented intent.

agentex/src/temporal/scheduled_agent_run_factory.py — per-fire engine allocation in `build_acp_use_case_for_principal` is worth revisiting if fire rates grow. agentex/src/adapters/temporal/adapter_temporal.py — string-based not-found detection is fragile to upstream message changes.
</details>

| Filename | Overview |
|----------|----------|
| agentex/src/domain/services/agent_run_schedule_service.py | Core service managing the dual Postgres+Temporal state. All previously flagged issues addressed. `trigger_schedule` uses `start_workflow` (not `trigger_now`), so manual fires won't update Temporal Schedule live stats. |
| agentex/src/temporal/activities/scheduled_agent_run_activities.py | Well-structured with idempotent task creation, delivered marker guard, fire-time authz re-check, and correct manual-trigger paused bypass. |
| agentex/src/adapters/temporal/adapter_temporal.py | Adds `update_schedule`; fixes `start_workflow` to use `args=` keyword. Not-found detection via string matching is fragile. |
| agentex/src/api/routes/agent_run_schedules.py | Clean route layer; `_check_schedule_or_collapse_to_404` imported with private prefix is a minor style concern. |
| agentex/src/temporal/scheduled_agent_run_factory.py | Per-fire factory creates a new DB engine on each activity invocation; worth revisiting under high fire rates. |
| agentex/src/api/schemas/agent_run_schedules.py | Schema validators correctly enforce single cadence at create and prevent both-non-null cadences at update. |
| agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py | Non-blocking schema-only migration; unique index enforces permanently reserved names consistently with application logic. |
| agentex/src/utils/schedule_metrics.py | Bounded categorical tags only — no high-cardinality entity IDs. |
| agentex/src/temporal/workflows/scheduled_agent_run_workflow.py | Thin deterministic wrapper; correctly passes trigger_type to the activity. |

</details>

<a href="#gh-light-mode-only">

```mermaid
sequenceDiagram
    participant API as API Route
    participant SVC as AgentRunScheduleService
    participant PG as Postgres
    participant TP as Temporal Schedule
    participant WF as ScheduledAgentRunWorkflow
    participant ACT as launch_scheduled_agent_run
    participant ACP as AgentsACPUseCase

    Note over API,TP: Create schedule
    API->>SVC: create_schedule(agent, request, creator_principal)
    SVC->>PG: create(AgentRunScheduleEntity)
    SVC->>TP: "create_schedule(temporal_id, args=[row.id])"
    TP-->>SVC: ScheduleHandle

    Note over TP,ACP: Each cron/interval fire
    TP->>WF: start ScheduledAgentRunWorkflow(schedule_id)
    WF->>ACT: launch_scheduled_agent_run(schedule_id, fire_id, trigger_type)
    ACT->>PG: get schedule row
    ACT->>ACT: fire-time authz re-check (creator_principal)
    ACT->>ACP: task/create (deterministic name)
    ACT->>ACP: message/send or event/send
    ACT->>PG: mark scheduled_input_delivered

    Note over API,TP: Manual trigger
    API->>SVC: trigger_schedule(agent_id, name)
    SVC->>TP: "start_workflow(ScheduledAgentRunWorkflow, args=[row.id, 'manual'])"
    TP-->>API: schedule response (async, no task info)
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
sequenceDiagram
    participant API as API Route
    participant SVC as AgentRunScheduleService
    participant PG as Postgres
    participant TP as Temporal Schedule
    participant WF as ScheduledAgentRunWorkflow
    participant ACT as launch_scheduled_agent_run
    participant ACP as AgentsACPUseCase

    Note over API,TP: Create schedule
    API->>SVC: create_schedule(agent, request, creator_principal)
    SVC->>PG: create(AgentRunScheduleEntity)
    SVC->>TP: "create_schedule(temporal_id, args=[row.id])"
    TP-->>SVC: ScheduleHandle

    Note over TP,ACP: Each cron/interval fire
    TP->>WF: start ScheduledAgentRunWorkflow(schedule_id)
    WF->>ACT: launch_scheduled_agent_run(schedule_id, fire_id, trigger_type)
    ACT->>PG: get schedule row
    ACT->>ACT: fire-time authz re-check (creator_principal)
    ACT->>ACP: task/create (deterministic name)
    ACT->>ACP: message/send or event/send
    ACT->>PG: mark scheduled_input_delivered

    Note over API,TP: Manual trigger
    API->>SVC: trigger_schedule(agent_id, name)
    SVC->>TP: "start_workflow(ScheduledAgentRunWorkflow, args=[row.id, 'manual'])"
    TP-->>API: schedule response (async, no task info)
```

</a>
</details>

<sub>Reviews (10): Last reviewed commit: ["fix(temporal): pass workflow args via ar..."](https://github.com/scaleapi/scale-agentex/commit/e6ea64b2c9d607061a6db2fd5356cb3b55fefb10) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39353769)</sub>


agentex/src/temporal/activities/scheduled_agent_run_activities.py
```diff
@@ -0,0 +1,308 @@
...
+            await use_case.handle_rpc_request(
+                method=AgentRPCMethod.EVENT_SEND,
+                params=SendEventRequestEntity(task_name=task_name, content=content),
+                agent_id=schedule.agent_id,
+            )
+
+        # Best-effort delivered marker for the retry guard above. A crash between
+        # delivery and this update is the only window where a retry could
+        # re-deliver; deterministic task naming still prevents duplicate tasks.
+        task.task_metadata = {
```
NiteshDhanpal, line 321: The deterministic task name prevents a duplicate task, but not a duplicate delivery right? event/send has no dedupe. For agentic agents that second event re-runs the whole turn (double reply, double LLM/tool cost etc..). We could add a idempotency key on the delivery so a replay is dropped

agentex/src/api/routes/agent_run_schedules.py
```diff
@@ -0,0 +1,247 @@
...
+    summary="Delete Run Schedule",
+    description="Delete a run schedule permanently.",
+)
+async def delete_run_schedule(
+    agent_id: str,
+    name: str,
+    run_schedules_use_case: DAgentRunSchedulesUseCase,
+    authorization: DAuthorizationService,
+) -> DeleteResponse:
+    await _check_schedule_or_collapse_to_404(
```
NiteshDhanpal, line 238: I think we should soft delete the schedule record for audit purposes

agentex/src/adapters/temporal/adapter_temporal.py
```diff
@@ -589,6 +616,70 @@ async def trigger_schedule(self, schedule_id: str) -> None:
...
+            await handle.update(_apply)
+            logger.info(f"Updated schedule {schedule_id}")
+        except Exception as e:
+            if "not found" in str(e).lower():
+                logger.error(f"Schedule {schedule_id} not found: {e}")
+                raise TemporalScheduleNotFoundError(
+                    message=f"Schedule '{schedule_id}' not found",
+                    detail=str(e),
+                ) from e
+            logger.error(f"Failed to update schedule {schedule_id}: {e}")
```
NiteshDhanpal, line 687: can we log a metric?

agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+    # Indexes target the table created in this same migration, so they hold no
+    # write-blocking lock against live traffic (the table has no rows yet).
+    op.create_index(
+        'uq_agent_run_schedules_agent_name',
+        'agent_run_schedules',
+        ['agent_id', 'name'],
+        unique=True,
+    )
+    op.create_index(
```
NiteshDhanpal, line 72: can we add an index on updated_at?

agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py
```diff
@@ -0,0 +1,81 @@
...
+# revision identifiers, used by Alembic.
+revision: str = '3b1c9d2e4f6a'
+down_revision: str | None = 'c7a1b2d3e4f5'
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        'agent_run_schedules',
```
NiteshDhanpal: Should we add a record version field to track every schedule update? Currently, patch, pause, and resume look like blind read-modify-write flows. That means a stale patch could accidentally overwrite a newer pause/resume change and silently reactivate a schedule.
If we add a version number, we can make updates conditional on the version the caller last read. That gives us optimistic concurrency control and also gives us a cleaner audit trail of schedule changes over time.

NiteshDhanpal (changes requested):

agentex/src/temporal/activities/scheduled_agent_run_activities.py
jromualdez-scale: Great point. You are right that the “correct” fix here is delivery-level idempotency in the existing Agentex send paths: add an optional idempotency_key to SendEventRequestEntity / SendMessageRequestEntity, persist it with the event/message, and enforce uniqueness so retries return/drop the existing delivery instead of forwarding again.

That would solve this cleanly, but it would also mean updating the existing Agentex event/message persistence paths, adding partial uniqueness indexes, and defining consistent behavior when an idempotency_key is provided. Existing normal client calls could keep the key optional and preserve current behavior, while scheduled-run retries would opt in with a deterministic key. We could also use this to address normal client retry idempotency more broadly if needed, but that feels like a bigger reliability effort than this v1 scheduling PR.

Another lightweight v1 option is to move the scheduled_input_delivered marker before delivery, so Temporal retries skip after the fire is claimed. But that flips the failure mode: we avoid rare double-delivery, at the cost of a rare missed delivery if the worker crashes after claiming and before sending.

Happy to go with either option here or keep it as is. Any thoughts @NiteshDhanpal? @danielmillerp any preference here?

agentex/src/api/routes/agent_run_schedules.py
jromualdez-scale: Agree this makes sense for audit. I think the shape would be to keep deleting the Temporal schedule so no future fires happen, but soft-delete the Postgres schedule row instead of removing it.

Implementation-wise we’d add a status/deleted_at field, filter deleted schedules out of list/get, and update delete_schedule to mark deleted after the Temporal delete succeeds. The main choice is whether deleted schedule names can be reused. Keeping the current unique (agent_id, name) means no reuse, which is simpler and preserves audit history cleanly. Allowing reuse would require a partial unique index over active schedules only.

Thoughts? It may be simpler in v1 to just assume deleted names cannot be reused

agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py
jromualdez-scale: I think this index would only help a query that sorts or filters on updated_at. Today the schedule list path orders by created_at, and I don’t see anything reading updated_at. It’s also the last-modified time of the schedule definition itself, which should only change on relatively rare patch/pause/resume operations.

Given the expected low row count per agent, I’d hold off until we have a query that actually needs it. Happy to add it if you’re picturing a “recently updated” sort/view down the line. Thoghts?

agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py
jromualdez-scale: Agreed that patch/pause/resume are blind read-modify-write flows today, and a version column would give us proper optimistic concurrency plus a cleaner foundation for change history.

My one hesitation is v1 scope: schedules are user-owned, and schedule mutations should be infrequent actions on a single row. In practice I’d expect one user to be editing/pausing/resuming a schedule one operation at a time, so concurrent conflicting edits seem unlikely compared to task/message/event traffic.

Do you think it’s worth addressing in this first pass, or could we defer the version column to a follow-up once we see real edit patterns?  Happy to go either way, just want to size it against v1 scope. cc @danielmillerp

agentex/src/adapters/temporal/adapter_temporal.py
jromualdez-scale: I'll add this!

agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py
NiteshDhanpal: My argument for doing this now is migration-cost asymmetry, not how often concurrency happens.
This table is brand new in the PR, so adding version INT NOT NULL DEFAULT 1 is basically free right now: no backfill, no nullable-then-populate migration, no schema change on a live populated table, and the write paths are already being built.
If we retrofit this later, we’ll need to migrate existing rows, backfill versions, and thread version checks through patch/pause/resume after those flows have already shipped. That is strictly more work and higher risk.
It also gives us a useful debugging hook for customer issues. If a customer reports an unexpected patch/pause/resume outcome, having an explicit version on the row makes it much easier to reason about which state was read, which update was attempted, and whether we dropped or overwrote a concurrent change.

agentex/src/api/routes/agent_run_schedules.py
```diff
@@ -0,0 +1,247 @@
...
+        authorization,
+        build_run_schedule_authz_selector(agent_id, name),
+        AuthorizedOperationType.update,
+    )
+    return await run_schedules_use_case.trigger_schedule(agent_id, name)
+
+
+@router.post(
+    "/{name}/pause",
```
danielmillerp, line 183: hell ya was gonna request haha

agentex/src/api/schemas/agent_run_schedules.py
```diff
@@ -0,0 +1,208 @@
...
+    is creator *context* used only for AuthZ and ownership at fire time.
+    """
+
+    principal_type: str | None = Field(
+        None, description="e.g. 'user' or 'service_account'."
+    )
+    user_id: str | None = Field(
+        None, description="Creator user id, if a user principal."
+    )
+    service_account_id: str | None = Field(
```
danielmillerp, line 41: this will accomodate for long term solution right?

agentex/src/domain/services/agent_run_schedule_service.py
```diff
@@ -0,0 +1,505 @@
...
+# and small (the row id is the only thing the workflow needs).
+RUN_SCHEDULE_TEMPORAL_ID_PREFIX = "agent-run-schedule"
+
+# Registered (class) name of the workflow each fire starts. Referenced by name so
+# the API/service layer doesn't import the Temporal workflow definition.
+SCHEDULED_AGENT_RUN_WORKFLOW_NAME = "ScheduledAgentRunWorkflow"
+
+
+def build_run_schedule_temporal_id(schedule_row_id: str) -> str:
+    return f"{RUN_SCHEDULE_TEMPORAL_ID_PREFIX}:{schedule_row_id}"
```
danielmillerp, line 48: we're using db to get this right?

agentex/src/domain/services/agent_run_schedule_service.py
```diff
@@ -0,0 +1,505 @@
...
+
+        try:
+            created = await self.schedule_repository.create(entity)
+        except DuplicateItemError as exc:
+            raise ClientError(
+                f"Run schedule '{request.name}' already exists for agent '{agent.id}'"
+            ) from exc
+
+        temporal_id = build_run_schedule_temporal_id(created.id)
+        authz_selector = build_run_schedule_authz_selector(agent.id, created.name)
```
danielmillerp, line 131: what is this

agentex/src/temporal/workflows/scheduled_agent_run_workflow.py
```diff
@@ -0,0 +1,39 @@
...
+    LAUNCH_SCHEDULED_AGENT_RUN_ACTIVITY,
+)
+from temporalio import workflow
+from temporalio.common import RetryPolicy
+
+
+@workflow.defn
+class ScheduledAgentRunWorkflow:
+    @workflow.run
+    async def run(self, schedule_id: str) -> dict[str, Any]:
```
danielmillerp: how long does this run, is it ever closed?

danielmillerp (approved):
overall looks great to me!

agentex/src/domain/services/agent_run_schedule_service.py
jromualdez-scale: yep!

agentex/src/temporal/workflows/scheduled_agent_run_workflow.py
jromualdez-scale: Each workflow execution is per scheduled fire and closes when the launch activity succeeds or exhausts retries. The long-lived thing is the Temporal Schedule, not this workflow.

agentex/src/api/schemas/agent_run_schedules.py
jromualdez-scale: Yep, and it's intentionally credential-free. We preserve the creator’s account/user/service-account context for ownership and attribution, but we do not store request credentials, headers, JWTs, API keys, etc. Worst case, even if the fire-time authz path evolves, this still gives us useful attribution/debugging context without retaining secrets.

agentex/src/domain/services/agent_run_schedule_service.py
jromualdez-scale: This is the AuthZ resource selector for the schedule. The general pattern in the codebase is AgentexResource(type, selector), and the selector is the stable id AuthZ uses for checks. Since schedule routes address schedules by (agent_id, name), we derive the selector from those path params so we can authorize before the DB lookup. It’s registered under the parent agent so schedule permissions can inherit/cascade from the agent.

I'll be honest Cursor figured this one out haha ^

agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py
danielmillerp: I'm aligned with having this be in table, but not implementing yet. That way we don't have to backfill like @NiteshDhanpal is saying in future. We can add functionality in future

agentex/src/api/routes/agent_run_schedules.py
NiteshDhanpal: Agree on the soft-delete approach — keep deleting the Temporal schedule so no future fires happen, and soft-delete the Postgres row after that delete succeeds, filtering deleted rows out of list/get
Fro the name reuse, I'm fine keeping the current (agent_id, name) unique index → no reuse for v1. It's the simplest path, and it actually keeps audit cleanest (one name = one schedule, ever). The only cost is you can't reuse a deleted name. If a real reuse need shows up later, a partial unique index (WHERE deleted_at IS NULL) is a clean, one-line upgrade that allows reuse

One bigger question to flag (non blocking for v1): the reuse debate only exists because name is our identity (URL + unique key + authz selector). The row already has a stable id; if id were the handle and name a mutable label, reuse would be a non-issue with unambiguous audit

agentex/src/temporal/activities/scheduled_agent_run_activities.py
NiteshDhanpal: I'd skip the marker-before option.. it trades the double delivery for a silent missed delivery, which is worse for a scheduled job. The current marker-after order gives us a clear at-least-once delivery model, which is the safer default

Can we just make that explicit in the code comments? i.e. state that delivery is at-least-once by design and the scenario when a double delivery can happen?

jromualdez-scale: Will do so thanks for the input here!

agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py
jromualdez-scale: Agreed will add this in!

agentex/src/temporal/activities/scheduled_agent_run_activities.py
danielmillerp: following discussion here! sorry for late reply here. Could pass in the ID through task metadata, and check on retry if that task metadata for the ID has already been sent to prevent duplicate fires? not a huge concern but just an idea

agentex/src/domain/services/agent_run_schedule_service.py
danielmillerp: yep as long as its consistent !

agentex/src/temporal/activities/scheduled_agent_run_activities.py
jromualdez-scale: @danielmillerp Thanks, that’s kinda the idea behind the current scheduled_input_delivered marker. The scheduled task is get-or-create by a deterministic name for the fire, then the activity checks task_metadata.scheduled_input_delivered before sending. After event/send or message/send succeeds, it writes that marker back onto the task so later retries skip delivery.

The remaining gap @NiteshDhanpal  is pointing out is the small window between delivery and the marker write. If the worker crashes after event/send/message_send succeeds but before scheduled_input_delivered is written, the retry sees no marker and can deliver the same initial input again. So task metadata gives us a best-effort guard, but it can’t fully guarantee message/event idempotency unless we either mark before send, which risks missed delivery, or add a real idempotency_key at the event/message persistence layer.

the fully correct fix would be follow-up idempotency support in event/send and message/send, but I think the current marker approach is sufficient for v1 and I can add the explicit comment as Nitesh reccomended above

agentex/src/adapters/temporal/adapter_temporal.py
jromualdez-scale: Added

agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py
jromualdez-scale: didn't make this addition in this round but this is an easy addition later on if needed. Let me know if you think otherwise

agentex/database/migrations/alembic/versions/2026_06_22_1200_add_agent_run_schedules_3b1c9d2e4f6a.py
jromualdez-scale: added

agentex/src/api/routes/agent_run_schedules.py
jromualdez-scale: Yeah, agreed. Using the row id as the external handle would be cleaner long-term: id would be the immutable resource identity, and name could just be a mutable/display label, which would make name reuse after soft delete a non-issue.

The reason this PR uses name is that the original API shape treated schedule name as the natural key (`/{name}`), and it lets us derive the AuthZ selector from path params before doing a DB lookup. With soft delete, I think the simplest v1-compatible choice is to keep `(agent_id, name)` unique forever, so one name maps to one schedule/audit trail.

I’d prefer not to switch the API identity to id in this PR unless we want to reopen the endpoint/AuthZ shape. We did have a similar issue with name uniqueness at the task level so this may be worth hammering down now in v1 to avoid rework later. @danielmillerp any preference here?

agentex/src/temporal/activities/scheduled_agent_run_activities.py
jromualdez-scale: Added the comment

agentex/src/api/routes/agent_run_schedules.py
danielmillerp: Agree, name shouldn't be the primary identity, but I don't want to bloat this PR's scope. Let's ship the soft-delete + (agent_id, name) unique / no-reuse here as you've laid out (soft-delete makes sense and no-reuse is fine for v1), and take the id-as-handle / name-as-mutable-label change as a fast-follow. It's a clean one since id is already returned and Temporal already keys off the row id, so moving the external handle to id later is additive rather than a breaking reshape.

jromualdez-scale: Sounds good! I'lll note it down in a comment so we remember to circle back to this

NiteshDhanpal (approved):

PR #343, opened 2026-06-30
docs: document Redis as a required dependency

vineetvora-scale:
## Summary

Redis is required at runtime for any agent that streams or sends messages, but the docs presented it as optional. Locally that's masked — the Docker stack starts Redis automatically — and in deployment `REDIS_URL` read like an optional custom secret. The result: an agent can pass local testing, deploy successfully, and then fail the first time it streams, with no clear signal that Redis was the missing piece.

This PR makes the docs tell the truth in every place a developer encounters Redis or credential mapping.

## What changed

| File | Change |
|------|--------|
| `README.md` | Prerequisites: Redis is required; the Docker stack starts it automatically (no local install); a local Redis conflicts on port 6379 |
| `CLAUDE.md` | Backend Services: mark Redis as required and auto-started, with the port-conflict note |
| `concepts/streaming.md` | Note that streaming and messaging are delivered through Redis (required) |
| `configuration.md` | Credentials Mapping: `REDIS_URL` is a required credential, not optional |
| `deployment/overview.md` | Add Redis / `REDIS_URL` to deployment prerequisites |
| `deployment/cicd.md` | Call out `REDIS_URL` as a required baseline secret; fix the misleading `redis://localhost/` example value |

## Scope

**Docs only — no code or template changes.** This PR documents the existing requirement; it does not change runtime behavior.

## Follow-ups (not in this PR)

- **SDK templates label `REDIS_URL` "Optional".** The scaffolding templates and example manifests in `scale-agentex-python` still present `REDIS_URL` as optional (and apply it inconsistently). They should be relabeled as required and made consistent.
- **Fail fast on unreachable Redis.** The SDK connects to Redis lazily and without a connect timeout, so a missing/unreachable Redis surfaces as a silent hang on first use rather than a clear startup error. A short connect timeout + an actionable error message would make the failure self-documenting. (Behavior change — needs SDK-owner sign-off.)
- **Provide `REDIS_URL` automatically at deploy.** Longer term, the platform could inject `REDIS_URL` the way it already provides other backing-service connectivity, so agent authors don't configure it by hand.

## Verification

`make build-docs` passes; all internal doc links resolve.

<h3>Greptile Summary</h3>

This is a docs-only PR that corrects the documentation to accurately reflect that Redis is a required runtime dependency for any agent that streams or sends messages, not an optional one. Previously, local development masked this requirement because the Docker stack auto-starts Redis, while `REDIS_URL` read as an optional custom secret in deployment docs.

[...]

PR #353, opened 2026-07-09
feat(dev): run agentex locally without Docker 

aringuyen3:
## What

Adds a **docker-free mode** that runs the full agentex backend as host processes with embedded datastores — a lighter alternative to the Docker Compose stack, closer to a one-command `langgraph dev`-style workflow.

```bash
./dev.sh no-docker                    # whole stack, no Docker
./dev.sh no-docker --lean             # Postgres + Redis + API + MongoDB only (no Temporal/OTel)
./dev.sh no-docker --no-temporal      # skip Temporal + the worker
./dev.sh no-docker --mongo-uri <uri>  # use an external MongoDB instead of a local mongod
```

The bare `./dev.sh` (Docker) is unchanged and now also accepts an explicit `./dev.sh docker` alias. The docker-free mode is also available as `make dev-no-docker` and `python -m scripts.dev_nodocker`.

## Why

Standing up a local environment previously required the full Docker stack. This lets a developer run the backend with a single command and no Docker daemon:

- Postgres via bundled `pgserver` (unix socket) and Redis via bundled `redislite`
- a Temporal dev server + Web UI and the agentex worker
- a local `mongod` — always started; the stack requires it (the Temporal worker builds Mongo-backed repositories at boot)
- an optional OpenTelemetry collector

It runs migrations, supervises uvicorn + the worker, and tears everything down cleanly on Ctrl-C / SIGTERM. `--ephemeral` uses a throwaway data dir; `--mongo-uri` points at an existing MongoDB instead of launching a local `mongod`.

## App-side change

Safe no-op wherever the backend runs in Docker / staging / prod (the env var it keys on is unset there):

**ACP host rewrite for docker-free mode.** Agents register their ACP URL at `host.docker.internal` (the SDK default, so a Docker backend can reach an agent on the host), which a host-process backend can't resolve. The runner sets `AGENTEX_ACP_HOST_OVERRIDE=127.0.0.1` and the backend rewrites only that sentinel host to the override when dialing agents — in the ACP request path, the agent-API-key proxy path, and the Temporal healthcheck. When the env var is unset, the stored URL is used verbatim. Default-scaffolded agents work without manifest edits.

## Also

- Fix a frontend dev-server process leak in `dev.sh` (kill the whole `make → npm → next` tree and sweep orphans; report status by listening port).
- Correct the MongoDB (`brew trust` first) and OpenTelemetry (release binary; not in Homebrew) install commands.
- Add a `dev-no-docker` uv dependency group (`pgserver`, `redislite`, `greenlet`) pulled only for docker-free mode.
- Document docker-free mode in `README.md` and `CLAUDE.md`.

## Contract note

The runner is a package (`scripts/dev_nodocker/`), so the direct invocation is `python -m scripts.dev_nodocker` (not `python scripts/dev_nodocker.py`). `./dev.sh no-docker` and `make dev-no-docker` are unchanged for callers.

## Testing

Manually exercised on macOS (Docker stopped):

- Full stack and `--lean` — the API passes `/healthz` within ~1s and `/readyz` reports Postgres, Redis, and MongoDB all healthy; the Temporal worker stays up; teardown frees all ports with no orphaned processes.
- Fail-fast path: full mode with `mongod` absent aborts with an actionable install message (or point at an external instance with `--mongo-uri`).
- Booted end-to-end via `python -m scripts.dev_nodocker`; agents scaffolded by `agentex init` connect without manifest edits (ACP host rewrite).

**Platform support:** macOS and Linux only. Not supported on native Windows — the embedded Redis (`redislite`) ships no Windows server build; run it under WSL2, where `redislite`/`pgserver`/`mongod` behave as on Linux (WSL2 path not yet verified).

<h3>Greptile Summary</h3>

This PR introduces a docker-free local development mode (`./dev.sh no-docker`) that provisions embedded datastores (bundled Postgres via `pgserver`, Redis via `redislite`, auto-downloaded Temporal dev server, and a local `mongod`) and supervises the API server and worker as host processes — no Docker daemon required. It also adds an ACP host-rewrite utility (`src/utils/acp_url.py`) so that agents registered at `host.docker.internal` are dialed correctly when the backend itself runs on the host.

- **New `scripts/dev_nodocker` package**: pure config module, service provisioners, subprocess supervisor, and orchestration runner. Previously flagged issues (crash returning exit code 0, empty-string `deployment.acp_url` falling through incorrectly) are both addressed in this revision.
- **ACP host rewrite (`resolve_acp_url`)**: applied consistently across all three ACP dial sites (ACP use case, API-key proxy, Temporal healthcheck activity). Safe no-op in Docker/staging/prod when `AGENTEX_ACP_HOST_OVERRIDE` is unset.
- **`dev.sh` enhancements**: `no-docker` subcommand, mode-file tracking for stop/status/restart, `kill_tree` + `sweep_stray_frontends` to fix the frontend process-leak bug, and a preflight port-conflict checker.

- Safe to merge — the new docker-free runner is additive and isolated behind an explicit subcommand; the ACP host-rewrite is a no-op in all non-local environments; and the two regressions called out in previous rounds are both corrected here.
- The orchestration, teardown, and mode-aware stop/status paths are well-structured. The ACP host-rewrite correctly covers all three dial sites and has no effect in Docker or production (env var absent). The empty-string deployment URL fallback and the crash exit-code were the two outstanding defects; both are fixed. The only remaining finding is that URL credentials are dropped by the netloc reconstructor — an edge case in a dev-tool path used exclusively with the local override.
- No files require special attention beyond the credential-preservation note in `agentex/src/utils/acp_url.py`.
</details>

| Filename | Overview |
|----------|----------|
| agentex/src/utils/acp_url.py | New utility that rewrites `host.docker.internal` to a local override in no-docker mode. Logic is clean and safe; minor issue with credentials being dropped when rewriting the netloc. |
| agentex/scripts/dev_nodocker/runner.py | Orchestration layer that provisions datastores, runs migrations, and supervises the API + worker. Previously flagged crash-exit-code bug (returning 0) is correctly fixed to return 1. Clean teardown in `finally` block. |
| agentex/scripts/dev_nodocker/services.py | Provisions embedded Postgres (pgserver), Redis (redislite), Temporal, MongoDB, and the OTel collector. Fail-fast on missing mongod with actionable install messages; OTel gracefully degrades. Teardown handles edge cases including the daemonized Redis PID. |
| agentex/scripts/dev_nodocker/supervise.py | Subprocess helpers: spawn with PIPE, TCP readiness probing, HTTP health polling, graceful SIGTERM → SIGKILL terminate. Correct `writer.wait_closed()` usage. |
| agentex/scripts/dev_nodocker/config.py | Pure config/env builder — no side effects, unit-testable. `build_env` properly removes env vars for optional services that weren't started. `resolve_config` covers ephemeral, custom data-dir, and default paths. |
| agentex/src/domain/use_cases/agents_acp_use_case.py | Previously flagged empty-string `deployment.acp_url` regression is correctly addressed — new code uses truthiness (`if not raw`) throughout, so an empty string from a deployment falls through to `agent.acp_url` instead of being dialed as `""`. `acp_url_override` also correctly passes through `resolve_acp_url`. |
| agentex/src/domain/use_cases/agent_api_keys_use_case.py | Applies `resolve_acp_url` to the agent-API-key proxy path so the host rewrite covers all three ACP dial sites. |
| agentex/src/temporal/activities/healthcheck_activities.py | Applies `resolve_acp_url` to the Temporal healthcheck activity, completing the trio of ACP dial sites that need the host rewrite. |
| dev.sh | Large but well-structured refactor. Adds `no-docker` subcommand with mode-file tracking, `preflight_port_check`, `kill_tree`, `sweep_stray_frontends`, and proper stop/status awareness. OTel binary installer proceeds without verification if the checksums file download fails (previously flagged P2, not yet addressed). |
| agentex/Makefile | Adds `dev-no-docker` target. Includes a TODO without a ticket number. |

</details>

```mermaid
sequenceDiagram
    participant devsh as dev.sh no-docker
    participant runner as runner.run()
    participant pg as pgserver (embedded)
    participant redis as redislite (embedded)
    participant mongo as mongod (local)
    participant temporal as Temporal dev server
    participant api as uvicorn API
    participant worker as Temporal worker

    devsh->>runner: asyncio.run(run(cfg))
    runner->>pg: provision_postgres(cfg)
    pg-->>runner: (server, database_url)
    runner->>redis: provision_redis(cfg)
    redis-->>runner: (server, redis_url)
    runner->>mongo: provision_mongo(cfg)
    mongo-->>runner: (proc, mongo_uri)
    runner->>temporal: provision_temporal(cfg)
    temporal-->>runner: (env, temporal_address)
    runner->>runner: build_env(...)
    runner->>runner: run_migrations(cfg, env)
    runner->>api: spawn("api", uvicorn ...)
    runner->>worker: spawn("worker", run_worker.py)
    runner->>runner: wait_for_health(/healthz)
    runner->>runner: asyncio.wait([stop, proc exits])
    Note over runner: Ctrl-C / SIGTERM sets stop
    runner->>worker: terminate(SIGTERM → SIGKILL)
    runner->>api: terminate(SIGTERM → SIGKILL)
    runner->>temporal: env.shutdown()
    runner->>mongo: terminate(SIGTERM → SIGKILL)
    runner->>redis: teardown_redis()
    runner->>pg: pg_server.cleanup()
```
</details>

<sub>Reviews (11): Last reviewed commit: ["Merge branch &#39;main&#39; into aringuyen/run-l..."](https://github.com/scaleapi/scale-agentex/commit/8b61bde391268a0eea0a5eb7491738847118b2ca) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43028738)</sub>


agentex/src/domain/use_cases/agents_acp_use_case.py
```diff
@@ -323,23 +324,29 @@ async def _resolve_acp_url(
...
-        # Resolve through production deployment if available
-        if agent.production_deployment_id:
+        The resolved URL is passed through resolve_acp_url so that, in docker-free
+        local mode, an agent registered at host.docker.internal is dialed at the
+        host-reachable override address (a no-op in Docker/prod).
+        """
+        raw = acp_url_override
+
+        # Prefer the production deployment's URL when there's no explicit override.
+        if raw is None and agent.production_deployment_id:
```
danielmillerp: @smoreinis can you take a look here just make sure it doesn't conflict with the preview workflow?

agentex/src/temporal/activities/healthcheck_activities.py
```diff
@@ -54,6 +55,9 @@ async def check_status_activity(self, agent_id: str, acp_url: str) -> bool:
         Returns:
             bool: True if the agent is healthy, False otherwise
         """
+        # In docker-free local mode, rewrite host.docker.internal -> the host-reachable
+        # override so the healthcheck matches how the request path dials the agent.
+        acp_url = resolve_acp_url(acp_url)
```
danielmillerp, line 60: just making sure, no other places we need to do conversion right?

agentex/src/domain/use_cases/agents_acp_use_case.py
```diff
@@ -323,23 +324,29 @@ async def _resolve_acp_url(
         agent: AgentEntity,
         acp_url_override: str | None = None,
     ) -> str:
-        """Resolve the ACP URL for an agent, optionally overriding with a specific URL."""
-        if acp_url_override:
-            return acp_url_override
+        """Resolve the ACP URL for an agent, optionally overriding with a specific URL.
```
danielmillerp, line 350: did we lose the override?

agentex/src/utils/acp_url.py
```diff
@@ -0,0 +1,56 @@
...
+_ACP_HOST_OVERRIDE_ENV = "AGENTEX_ACP_HOST_OVERRIDE"
+
+
+def resolve_acp_url(url: str) -> str:
+    """Rewrite a ``host.docker.internal`` ACP host to the local override, if set.
+
+    Returns the URL unchanged when the override env var is unset or the URL does
+    not use the Docker sentinel host, so it is safe to call on every ACP dial.
+    """
+    override = os.environ.get(_ACP_HOST_OVERRIDE_ENV)
```
danielmillerp, line 41: is this where the override moved?

agentex/Makefile
danielmillerp, line 40: these are both technically local maybe rename to something more meaningful?

CLAUDE.md
```diff
@@ -55,6 +55,27 @@ Other commands:
...
+Docker-free local mode (host processes + embedded datastores, no Docker):
+```bash
+./dev.sh local              # whole stack without Docker (or: make dev-local)
+./dev.sh local --lean       # Postgres + Redis + API + MongoDB only
+./dev.sh local --mongo-uri <uri>  # use an external MongoDB instead of a local mongod
+```
+
+> **MongoDB is required for the full local stack** and is always started — the Temporal
+> worker builds Mongo-backed repositories at startup, so a missing/unreachable Mongo
+> makes the runner fail fast (with an install message) rather than crash the worker.
```
danielmillerp, line 67: ty for adding here

danielmillerp (commented):
lots of small questions! also as far as testing on PC, not super sure. A lot of our clients do use PCs and I know there are Scaliens who have PCs. What problems do you anticipate?

levilentz (commented):
Overall this is rad. 

One UX suggestion: in addition dev.sh, I think this would be really nice to have in the agentex CLI. I.E. you can run a single agent inline. therefore to start running a single agent you can start quick rather than having both the agentex backend and cli running. then this setup would be suitable for someone testing a multiagent flow. 

Not sure if that is possible given the abstraction and integration, but wanted to throw it out there as a north star.

danielmillerp:
> Overall this is rad.
> 
> One UX suggestion: in addition dev.sh, I think this would be really nice to have in the agentex CLI. I.E. you can run a single agent inline. therefore to start running a single agent you can start quick rather than having both the agentex backend and cli running. then this setup would be suitable for someone testing a multiagent flow.
> 
> Not sure if that is possible given the abstraction and integration, but wanted to throw it out there as a north star.

+1 to that!

aringuyen3:
> lots of small questions! also as far as testing on PC, not super sure. A lot of our clients do use PCs and I know there are Scaliens who have PCs. What problems do you anticipate?

@danielmillerp So Im testing this on Windows (via AWS Workspaces) but the embedded Redis uses (the [redislite package ](https://pypi.org/project/redislite/) has no native Windows build (Redis ships no supported Windows server) so the dependency required to run locally without Docker won't even install on native Windows.

agentex/src/temporal/activities/healthcheck_activities.py
aringuyen3: yes, only running this mode

agentex/src/utils/acp_url.py
aringuyen3: yes, here we override the environment variable `_ACP_HOST_OVERRIDE_ENV`. This environment is set and passed from `config.py` file

agentex/src/domain/use_cases/agents_acp_use_case.py
aringuyen3: hi, so ./dev.sh no-docker will pass _ACP_HOST_OVERRIDE_ENV variable, so the function resolve_acp_url will resolve this. Besides this, the logic  of this function is kept the same

aringuyen3:
### Some findings from testing local dev on Windows:

- I'm testing running locally on native Windows, but the embedded Redis uses the [redislite](https://pypi.org/project/redislite/) package, which has no native Windows build (Redis ships no supported Windows server). So the dependency required to run locally won't even install on native Windows.
- Workarounds today: run dev locally with Docker, or use the Docker-free local mode inside WSL2.
- Open question for FDE / customer environments: can they install WSL2 in their env? I wouldn't assume it's always available.
- Worth noting: I use AWS Workspace, which doesn't support WSL2 — so even the WSL2 path isn't universal internally.

@NiteshDhanpal had a clear view on how to scope this. His take: don't treat Windows-native local mode as a hard requirement for now (given the redislite/Redis limitation), and use a support matrix instead:

- **Mac/Linux** → Docker-free local mode
- **Windows with WSL2 allowed** → Docker-free local mode inside WSL2
- **Windows without WSL2 / locked-down envs** → Docker-based local dev

His main concern is a reliable fallback so customers aren't blocked when local setup is painful or impossible, and he thinks a cloud/dev-environment fallback is the most reliable option for that.

danielmillerp:
> ### Some findings from testing local dev on Windows:
> * I'm testing running locally on native Windows, but the embedded Redis uses the [redislite](https://pypi.org/project/redislite/) package, which has no native Windows build (Redis ships no supported Windows server). So the dependency required to run locally won't even install on native Windows.
> * Workarounds today: run dev locally with Docker, or use the Docker-free local mode inside WSL2.
> * Open question for FDE / customer environments: can they install WSL2 in their env? I wouldn't assume it's always available.
> * Worth noting: I use AWS Workspace, which doesn't support WSL2 — so even the WSL2 path isn't universal internally.
> 
> @NiteshDhanpal had a clear view on how to scope this. His take: don't treat Windows-native local mode as a hard requirement for now (given the redislite/Redis limitation), and use a support matrix instead:
> 
> * **Mac/Linux** → Docker-free local mode
> * **Windows with WSL2 allowed** → Docker-free local mode inside WSL2
> * **Windows without WSL2 / locked-down envs** → Docker-based local dev
> 
> His main concern is a reliable fallback so customers aren't blocked when local setup is painful or impossible, and he thinks a cloud/dev-environment fallback is the most reliable option for that.

I'm aligned @levilentz does that make sense for customers? Also CC: @lucyakoroleva for context per today's standup discussion

danielmillerp (approved):

levilentz:
> I'm aligned @levilentz does that make sense for customers? Also CC: @lucyakoroleva for context per today's standup discussion

@danielmillerp  @aringuyen3 I think broadly this is a good first step. I think assuming that the customer will not have WSL or docker is the way to test this as that it normally the problem. 

can we just disable redis in this mode?

danielmillerp:
> > I'm aligned @levilentz does that make sense for customers? Also CC: @lucyakoroleva for context per today's standup discussion
> 
> @danielmillerp @aringuyen3 I think broadly this is a good first step. I think assuming that the customer will not have WSL or docker is the way to test this as that it normally the problem.
> 
> can we just disable redis in this mode?

I like that! not necessary to go from 0 to 1 to stream

aringuyen3:
> > > I'm aligned @levilentz does that make sense for customers? Also CC: @lucyakoroleva for context per today's standup discussion
> > 
> > 
> > @danielmillerp @aringuyen3 I think broadly this is a good first step. I think assuming that the customer will not have WSL or docker is the way to test this as that it normally the problem.
> > can we just disable redis in this mode?
> 
> I like that! not necessary to go from 0 to 1 to stream

@danielmillerp @levilentz Redis is required. Agentex uses Redis for streaming and messaging. Referred to this PR https://github.com/scaleapi/scale-agentex/pull/343

PR #365, opened 2026-07-16
feat(tasks): non-terminal task/interrupt + INTERRUPTED status (AGX1-391)

declan-scale:
## Summary

Platform half of AGX1-391 (stop an agent run mid-stream). Adds a non-terminal `INTERRUPTED` task status and a `task/interrupt` operation so a running turn can be stopped without terminating the task or locking the chat.

## Ownership model (important)

Interrupt is **cooperative**, not control-plane-authoritative like `cancel`:

- **`task/interrupt` RPC = courier only.** It is forwarded to the agent (exactly like `event/send`) and writes no status. Only the agent can actually stop an in-flight turn.
- **The agent records `INTERRUPTED`.** After stopping its turn, the agent calls the REST **`POST /tasks/{task_id}/interrupt`** route (like the other task-state routes: complete/fail/cancel/timeout) to set its status. An agent that does not implement interrupt simply stays `RUNNING`, so status stays honest.
- **The control plane owns `RUNNING`.** It sets `RUNNING` on task creation and auto-resumes `INTERRUPTED -> RUNNING` on the next turn (`_get_or_create_task`).

Contrast with `cancel`, which is authoritative: the control plane sets `CANCELED` itself (and the RPC `cancel_task` also CAS-writes `CANCELED`), because cancel does not depend on the agent cooperating.

## What's included

- `task/interrupt` RPC method (forward-only) + REST `POST /tasks/{task_id}/interrupt` (authorized `update`, the agent's write path). Allowed for `SYNC` / `ASYNC` / `AGENTIC`.
- New non-terminal `INTERRUPTED` status + alembic migration; terminal transitions accept `INTERRUPTED` as a valid source (so an interrupted task can still be canceled/completed).
- Auto-resume `INTERRUPTED -> RUNNING` on the next turn; if that CAS loses to a concurrent terminal transition, the turn is rejected instead of running against a terminal task.
- Regenerated `openapi.yaml` (drives SDK regeneration).
- Unit + integration tests mirroring the `task/cancel` suite.

## Review follow-ups (in this PR)

- `task/interrupt` RPC reworked to forward-only per design discussion (agent owns `INTERRUPTED`); control plane keeps ownership of `RUNNING`/auto-resume.
- Auto-resume rejects with `ClientError` if the resume CAS lost to a concurrent cancel/complete (@smoreinis).
- `cancel_task` (RPC path) hardened with a compare-and-swap so it cannot clobber a task the agent terminated during the ACP call (independent correctness fix; @greptile).
[...]

PR #366, opened 2026-07-16
feat(ui): Stop button + queue-at-boundary for task interrupt (AGX1-391)

declan-scale:
Draft. Frontend half of AGX1-391 for the native agentex chat. Split out of the backend PR scaleapi/scale-agentex#365.

## Included
- Dual-mode composer button: **Stop** when a turn is streaming and the input is empty; **Send** (which enqueues) when there is text; normal Send when idle.
- Esc-to-interrupt while streaming.
- Queued-message chip, flushed on the streaming->idle edge (queue-at-boundary).
- `INTERRUPTED` treated as non-terminal (composer stays enabled; running is judged from `rpcStatus`).
- Sync `message/send` AbortController exposed so Stop tears the stream down; meta cache settled to `idle` so no hung 'thinking' indicator.

## Depends on
- scale-agentex#365 (the `POST /tasks/{id}/interrupt` route + `INTERRUPTED` status). The interrupt call uses a thin REST adapter with a `TODO(AGX1-391)` to swap to the typed SDK method after regeneration.

## Note
- A couple of explanatory comments were added in the `.ts` hooks; will strip to match the no-inline-comment convention before marking ready.

### DIFF
--- a/agentex-ui/components/primary-content/prompt-input.tsx
+++ b/agentex-ui/components/primary-content/prompt-input.tsx
@@ -8,7 +8,7 @@ import { Prec } from '@codemirror/state';
 import { EditorView, keymap } from '@codemirror/view';
 import CodeMirror from '@uiw/react-codemirror';
 import { DataContent, TextContent } from 'agentex/resources';
-import { ArrowUp } from 'lucide-react';
+import { ArrowUp, Square, X } from 'lucide-react';
 
 import { useAgentexClient } from '@/components/providers';
 import { IconButton } from '@/components/ui/icon-button';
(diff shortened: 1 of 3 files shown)


declan-scale:
Closing per design discussion on AGX1-391: interruption must be implemented by the agent itself (the agent stops its own turn and sets its state to INTERRUPTED; the control plane just forwards `task/interrupt` like `event/send`). A generic Stop button in the native agentex-ui would be misleading for any agent that does not implement self-interruption. The golden-agent-specific Stop + queue-at-boundary UI stays in scaleapi#151564 since the golden agent implements the interrupt handler.

PR #411, opened 2026-08-27
feat(agentex): filter GET /agents by agent card metadata

declan-scale:
:trophy: Brought to you by the Golden Agent ([Try it out](https://app.notion.com/p/Getting-Started-with-Golden-Agent-381904d6e6cb81adb8a8fbdbdd2f0bc8))

## Problem

`AgentCard` publishes self-description data through `registration_metadata.agent_card`, but `GET /agents` has no way to filter on its contents. Discovery flows that want to enumerate agents opting into a specific protocol capability (e.g. Permits' workflow submission protocol) have no server-side hook and must fetch everything client-side.

Linear: [AGX1-1048](https://linear.app/scale-epd/issue/AGX1-1048/add-agentcard-metadata-field-and-filter-agents-by-card-metadata)

## Change

* Adds an optional `agent_card_metadata` query parameter to `GET /agents`. The value is a JSON-encoded object; malformed JSON or non-object payloads return `400`.
* The route parses the JSON and forwards it to `AgentsUseCase.list`, which reserves the key `agent_card_metadata` in the repository filters dict.
* `AgentRepository.list` applies a JSONB `@>` filter at the top level:
  ```sql
  registration_metadata @> jsonb_build_object('agent_card', jsonb_build_object('metadata', :value))
  ```
  Wrapping under the same nested shape as the stored card means agents whose `registration_metadata` is `NULL`, missing `agent_card`, or missing `agent_card.metadata` are naturally excluded, and every requested key/value must be present at the correct nesting level.
* No DB migration — `agents.registration_metadata` is already `JSONB`.
* Existing pagination, ordering, task filtering and authorization behavior are preserved (the filter composes with the pre-existing `task_id` join, authorization id set, and `status != DELETED` clause).
* `openapi.yaml` regenerated by hand to reflect the new query parameter; the paired SDK PR consumes the same spec.

## Test coverage added

* Integration tests in `tests/integration/api/agents/test_agents_api.py`:
  * Matching key/value returns only opted-in agents; agents with different values or no card at all are excluded.
  * Non-matching value returns an empty list.
  * Multi-key filter requires containment of every key/value.
  * The filter composes with `limit`/`page_number`.
  * Malformed JSON and non-object payloads return `400`.
* Unit tests in `tests/unit/use_cases/test_agents_use_case.py` that seed agents directly via the repository and exercise the use-case-to-repo plumbing against real Postgres (single-key, multi-key, absent-card, and omitted-filter cases). Tests use a per-invocation tag so they are safe against session-scoped container reuse.

## Test plan (for reviewer, since local yarn/uv installs are skipped per Golden Agent policy)

- [ ] CI unit tests (`make test-unit`) pass, including the two new use-case tests.
- [ ] CI integration tests (`make test-integration`) pass, including the four new API tests.
- [ ] Manually curl `GET /agents?agent_card_metadata={\"permits_capable\":true}` against a dev backend seeded with an agent card and confirm only that agent is returned.
- [ ] Confirm `openapi.yaml` still matches the FastAPI-generated spec (`make gen-openapi` should produce no further diff).

## Out of scope / follow-ups

* Range queries, arbitrary operators, or a query language beyond exact containment.
* Deployment-scoped cards or resolving card metadata through `production_deployment_id`.
* Performance indexing — the endpoint uses top-level containment on an already-JSONB column and current agent counts don't warrant a GIN index yet. Revisit if listing at scale becomes a bottleneck.

<h2><a href="https://app.greptile.com/api/retrigger?id=57553867"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge.

<h3>Summary</h3>

- The route parses the JSON query value and returns `400` for malformed or non-object input.
- The use case passes the filter to the repository, which applies nested JSONB containment while preserving existing filters and pagination.
- A concurrent GIN index supports the new lookup, and API and use-case tests cover matching, pagination, invalid input, and card updates.

<details><summary><h3>Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant Client
    participant Route as GET /agents
    participant UseCase as AgentsUseCase.list
    participant Repo as AgentRepository.list
    participant DB as PostgreSQL
    Client->>Route: agent_card_metadata query value
    alt Invalid JSON or non-object
        Route-->>Client: 400
    else Valid object or omitted
        Route->>UseCase: list(agent_card_metadata)
        UseCase->>Repo: list(filters)
        Repo->>DB: SELECT non-deleted authorized agents
        opt Filter was supplied
            Repo->>DB: "registration_metadata @> nested agent_card metadata"
        end
        DB-->>Repo: matching page
        Repo-->>UseCase: agents
        UseCase-->>Route: agents
        Route-->>Client: 200 agent list
    end
```
</details>

<sub>Reviews (8) · Last reviewed commit: ["docs(agentex): note the deployment-path ..."](https://github.com/scaleapi/scale-agentex/commit/2a975e823afb0f6adcacb697a150bbb43f35706b)</sub>


basselatscale (changes requested):
The JSONB containment direction is right, and the integration coverage proves the important subset behavior: an agent whose card contains additional metadata still matches {"permits_capable": true}.

A few changes are needed before this is ready:

1. list_agents() now fails when called directly without agent_card_metadata. Its default is a FastAPI Query object, so json.loads(agent_card_metadata) raises TypeError. This is currently failing the two authorization unit tests. Please use the Annotated[..., Query(...)] = None form, or otherwise ensure the Python default is actually None, and keep the direct-call tests passing.

2. In AgentRepository.list, use if agent_card_metadata is not None: rather than a truthiness check. Otherwise an explicitly supplied {} silently bypasses the metadata predicate and includes agents with missing metadata.

3. Please ensure the OpenAPI/SDK contract supports an ergonomic mapping input rather than requiring every caller to manually json.dumps it. The required consumer shape is:

    client.agents.list(
        agent_card_metadata={"permits_capable": True},
    )

If the wire parameter must remain JSON encoded, the generated/client layer should perform that encoding. The current string schema generates string-typed SDK parameters.

Once those are fixed, this server-side capability is sufficient for our immediate goal: discovering AgentCard-published workflow descriptors and removing the generated input-contract bundle.

declan-scale:
Addressed all three review items in 52b3277:

1. **Direct-call default**: `agent_card_metadata` now uses `Annotated[str | None, Query(...)] = None`, so calling `list_agents()` directly gets a real `None` default. The two authz unit tests pass again (updated to expect the forwarded `agent_card_metadata=None` kwarg).
2. **Empty-object bypass**: the repository now applies the containment predicate on `is not None`, so an explicit `{}` requires a card metadata object to be present. Added an integration test covering this.
3. **SDK contract**: the query parameter is now declared in the spec with `content: application/json` and an `object` schema (via `openapi_extra`; the runtime string param is schema-hidden). That's the OpenAPI signal for a mapping-typed SDK parameter that the client JSON-encodes on the wire, so `client.agents.list(agent_card_metadata={"permits_capable": True})` becomes the generated shape. `openapi.yaml` regenerated via `make gen-openapi`.

basselatscale (commented):
Follow-up review — all three original items are addressed ✓

Two new warnings worth confirming before merge:

**1. Non-finite JSON values in `agent_card_metadata`**
Python's `json.loads()` accepts `NaN`, `Infinity`, and extreme exponents like `1e1000000` which aren't valid interoperable JSON. These will pass the current `isinstance(parsed, dict)` check but can fail at the PostgreSQL JSONB binding layer with an uncontrolled 500 instead of a clean 400. Low likelihood in practice, but a defense-in-depth gap. Could be a follow-up — e.g. `json.loads(agent_card_metadata, parse_constant=lambda _: None)` + a `math.isfinite` walk, or just `parse_float=decimal.Decimal`.

**2. Stainless SDK `Parameter/MissingSchema` warning**
The Stainless bot reports "Defaulted parameter to `type: string` because no schema was defined" on all three SDK previews. Can you confirm the generated Python and TypeScript SDK type signatures actually expose `agent_card_metadata` as a `dict`/`object` (not `str`)? If the preview SDKs are correct the warning is cosmetic, but if they defaulted to `str` it defeats the intent of the `content: application/json` spec encoding.

declan-scale:
Both follow-up warnings addressed in 1a4c251.

**1. Non-finite JSON values — fixed.**

`json.loads` now runs with `parse_constant` and `parse_float` hooks that reject anything that isn't interoperable JSON, so these all return a clean `400` instead of reaching the JSONB bind parameter:

| input | before | after |
|---|---|---|
| `{"x": NaN}` | 500 | 400 |
| `{"x": Infinity}` / `{"x": -Infinity}` | 500 | 400 |
| `{"x": 1e1000000}` (overflows to `inf`) | 500 | 400 |
| `{"x": <5000-digit int>}` | 500 | 400 |

Nested cases (`{"x": [1, NaN]}`, `{"x": {"nested": Infinity}}`) are covered too, since the hooks fire at every level. Added as a parametrized integration test. Note the same latent gap exists on `GET /tasks?task_metadata=`, which parses with a bare `json.loads` — happy to fix that in a follow-up rather than widen this PR.

**2. Stainless `Parameter/MissingSchema` — not cosmetic. You were right to ask.**

I checked the preview build directly rather than assuming:

```python
# stainless-sdks/agentex-sdk-python @ preview/.../agx1-1048-agent-card-metadata-filter
# src/agentex/types/agent_list_params.py
class AgentListParams(TypedDict, total=False):
    agent_card_metadata: str
```

So the generator ignores `content: application/json` on a query parameter entirely — that's exactly what the warning was reporting — and defaults to `type: string`. The mapping-typed parameter never materialized, and the `openapi_extra` block was buying nothing while costing a build regression.

I've dropped it. `agent_card_metadata` is now declared the same way the already-shipped `GET /tasks?task_metadata=` filter is declared: a nullable string carrying a JSON-encoded object.

```yaml
- name: agent_card_metadata
  in: query
  required: false
  schema:
    anyOf:
      - type: string
      - type: 'null'
    description: 'JSON-encoded object used to filter agents on ...'
```

That clears the warning, makes the two containment filters consistent, and means the generated client surface is reproducible from the spec with no hand-editing.

The ergonomics you asked for now live in the SDK's hand-written layer instead of fighting the generator — see the paired SDK PR, which adds `encode_metadata_filter`:

```python
from agentex.lib.utils.metadata_filters import encode_metadata_filter

client.agents.list(
    agent_card_metadata=encode_metadata_filter({"permits_capable": True}),
)
```

It's one call rather than a bare `json.dumps`, it pins key order, and it rejects `NaN`/`Infinity` client-side so you get a clear local error instead of a round-trip 400. If you'd rather have `client.agents.list(agent_card_metadata={...})` literally, the only way to get there reproducibly is to move the filter into a request body (a `POST /agents/search`) — worth doing if we expect this filter surface to grow, but it's a bigger change than this PR and I'd rather do it deliberately than smuggle it in here.

**Tests:** 26 passed in `tests/integration/api/agents/test_agents_api.py`, 22 in `tests/unit/api/test_agents_authz.py`, 8 in `tests/unit/use_cases/test_agents_use_case.py`. `make gen-openapi` produces no further diff.

basselatscale (changes requested):
The filtering behavior and validation changes look good, and the local Permits vertical-slice test passed against this exact head.

One production concern remains before we adopt this as a polled workflow registry: the new containment query runs against `agents.registration_metadata`, but the PR does not add a supporting JSONB index. Every Permits harness replica will periodically page through this filtered endpoint, while agent registrations are comparatively write-light. Without an index, that creates a recurring full scan as the registry grows.

Please add a backward-compatible Alembic migration for a concurrent GIN index matching the current predicate, for example:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS ix_agents_registration_metadata_gin
ON agents
USING gin (registration_metadata jsonb_path_ops);
```

Please use Alembic's autocommit support because `CREATE INDEX CONCURRENTLY` cannot run inside the normal migration transaction, and make the downgrade concurrent as well.

I recommend indexing the full `registration_metadata` expression used by the current query rather than introducing an expression/partial index in this PR. That keeps the implementation aligned with the existing `registration_metadata @> ...` predicate and avoids an API or repository-query redesign.

Please also add the normal migration-level verification used by this repository. No Redis caching, registry revision key, or pagination change is requested here; Permits will own its refresh interval, jitter, last-known-good snapshot, and submission provenance.

basselatscale (changes requested):
One additional regression case would make the paired SDK worker change safe to rely on.

The SDK follow-up will pass an optional AgentCard through `AgentexWorker`'s existing automatic `/agents/register` call rather than issuing a second registration. Please add an integration test proving idempotent re-registration of the same agent replaces the stored top-level `registration_metadata.agent_card` and immediately changes metadata-filter results:

1. register the agent with card metadata A;
2. verify filter A matches;
3. re-register the same agent identity with card metadata B;
4. verify filter A no longer matches and filter B does.

The current top-level `existing_metadata.update(registration_metadata)` implementation appears to provide this behavior already, so this should primarily lock the release-update contract rather than require an API redesign.

Please also document or test the supported way to withdraw a previously published AgentCard during rollback. Omitting `registration_metadata` currently preserves existing metadata, which can leave a stale discoverable descriptor after reverting to a release that no longer publishes a card. An explicit `{"agent_card": null}` registration may already provide the required non-breaking clearing behavior; if that is the intended contract, please lock it with a test.

This is separate from the previously requested GIN migration, and does not require registry webhooks or a revision API.

declan-scale:
Addressed both reviews in d801b33.

**GIN index migration — added.** `2026_09_03_1200_add_agents_registration_metadata_gin_index_b7d3e1f4a2c6.py` creates `ix_agents_registration_metadata_gin` on the full `registration_metadata` column with `jsonb_path_ops`, exactly matching the `@>` predicate the repository emits. The upgrade and the downgrade both run `CONCURRENTLY` inside `autocommit_block`, with `IF NOT EXISTS` / `IF EXISTS` guards. The repo migration linter (`migration_lint.py`) passes. I also verified the migration against a throwaway Postgres 16 container: `alembic upgrade head` creates the index with the expected definition, and `alembic downgrade -1` removes it.

**Re-registration contract — locked with tests.**
- `test_reregistration_replaces_agent_card_and_filter_results`: register with card A, filter A matches; re-register the same agent with card B, filter A returns empty and filter B matches.
- `test_reregistration_with_null_agent_card_withdraws_from_discovery`: an explicit `{"agent_card": null}` registration clears the published card, so filter and `{}` filter both stop matching while the agent itself stays listed. The test also proves that omitting `registration_metadata` preserves the existing card, so the clearing is opt-in and non-breaking.

Both behaviors come from the existing top-level `existing_metadata.update(...)` merge (`agents_use_case.py:162-165`); no API change was needed. Full `test_agents_api.py` suite: 28 passed.

basselatscale (approved):
LGTM

agentex/src/api/routes/agents.py
```diff
@@ -107,6 +110,57 @@ async def get_agent_by_name(
     return Agent.model_validate(agent_entity)
 
 
+_AGENT_CARD_METADATA_DESCRIPTION = (
```
jromualdez-scale, line 113: Follow-up, not for this PR: registrations that carry a `deployment_id` go through `complete_deployment_registration`, which writes to the deployment row and never touches `agents.registration_metadata`. So this filter only sees cards published via the legacy (non-deployment) registration path. Fine for Permits, whose workers register the legacy way, but the next team to try this with the cloud-build path will get an empty list and no hint why. A sentence here would save them the debugging.

Suggesting a follow-up rather than a change now because this constant feeds the OpenAPI spec, and #502's generated docstrings were verified byte-identical to the Stainless preview. Changing it here means re-touching #502, which isn't worth holding a customer-blocked merge for.

jromualdez-scale (approved):
Reviewed from the platform side. The containment predicate, the `is not None` handling for `{}`, the strict JSON parsing, and the concurrent GIN migration all look right, and the re-registration / withdrawal tests are the right contract to lock. Verified the migration chains off the current head on `main` and mirrors the existing `ix_tasks_metadata_gin` pattern.

Approving. One follow-up inline, plus +1 to the `GET /tasks?task_metadata=` hardening follow-up you mentioned. When you file that one, worth deciding in the same ticket whether `{}` should mean the same thing on both filters: tasks currently rejects it with a 400 and agents treats it as "has a metadata object". Both defensible, but they're the only two JSON-containment filters in the API and they now diverge.

agentex/src/api/routes/agents.py
declan-scale: Done in this PR (f60572c) rather than a follow-up, so the pair ships as a complete e2e solution. The query parameter description now states that only direct-path registrations match, and that a deployment_id registration writes the card to the deployment record. openapi.yaml is regenerated, and #411's paired SDK PR syncs the generated docstrings to the same text.

